PR #5, opened 2019-10-30
Empty transliteration search return empty string implemented

yCobanoglu:


ebl/fragmentarium/domain/transliteration_query.py
```diff
@@ -29,6 +29,14 @@ def regexp(self):
         )
         return fr'{lines_regexp}(?![^|\s])'
 
+    def is_sequence_empty(self) -> bool:
+        if all(list(map(lambda x: not x, self._signs))):
+            return True
+        elif all(list(map(lambda x: x == '', self._signs[0]))):
```
jlaasonen: This will fail for e.g. `[[''], ['ABZ001']]`.

The if-else-block could be replaced by a single check like:

```
''.join(token for row in sequence for token in row).trim() == ''
``

File: ebl/tests/fragmentarium/test_fragment_finder.py [hunk 2]
```diff
@@ -87,6 +89,13 @@ def test_search(fragment_finder, fragment_repository, when):
     assert fragment_finder.search(query) == [FragmentInfo.of(fragment)]
 
 
+GET_SEARCH_TRANSLITERATION_EMPTY_DATA = [
```
jlaasonen: This is far from the point of use. It should be moved next to `test_search_transliteration_empty`.  It is also possible to inline it to `@pytest.mark.parametrize`.

ebl/fragmentarium/domain/transliteration_query.py
```diff
@@ -29,6 +29,14 @@ def regexp(self):
         )
         return fr'{lines_regexp}(?![^|\s])'
 
+    def is_sequence_empty(self) -> bool:
```
jlaasonen: The name could be just `is_empty`. Using the `Sequence` type is an implementation detail.

ebl/fragmentarium/domain/transliteration_query.py
```diff
@@ -29,6 +29,16 @@ def regexp(self):
...
+        if all(list(map(lambda x: not x, self._signs))):
+            return True
+        elif ''.join(
```
jlaasonen: This check should be enough. There is no need for the if-block.

PR #14, opened 2020-02-07
Tokenize inside reading names

jlaasonen:
See: https://trello.com/c/gmQjHa6D


jlaasonen:
[Parse words](https://trello.com/c/gmQjHa6D/343-parse-words)

PR #21, opened 2020-03-19
fix error

yCobanoglu:


jlaasonen:
[VisitError: Error trying to process rule "ebl_atf_at_line__edge":](https://trello.com/c/WVo2agCf/386-visiterror-error-trying-to-process-rule-eblatfatlineedge)

PR #22, opened 2020-03-23
fix VisitError Bug

yCobanoglu:


jlaasonen:
[VisitError: Error trying to process rule "ebl_atf_at_line__column":](https://trello.com/c/j9G4Eodq/389-visiterror-error-trying-to-process-rule-eblatfatlinecolumn)

ebl/transliteration/domain/lark_parser.py
```diff
@@ -96,6 +96,14 @@ def parse_line_(line: str, line_number: int):
...
                 {"description": f"Invalid brackets.", "lineNumber": line_number + 1,},
             )
+        except VisitError as ex:
+            return (
+                None,
+                {
+                    "description": f"Invalid Value: {ex.orig_exc}",
+                    "lineNumber": line_number + 1,
+                },
+            )
```
jlaasonen: Not all `VisitError`s are due to an invalid value. You should only handle the specific case, other errors should still lead to an internal server error.

yCobanoglu:
> Add test cases for error handling.

To test the VisitError i would have to do integration testing. Currently only **Duplicate Status** is tested

ebl/transliteration/domain/labels.py
```diff
@@ -10,6 +10,11 @@
 from ebl.transliteration.domain.atf import Status, Surface
 
 
+class DuplicateStatusError(ValueError):
+    def __init__(self, msg):
+        super().__init__(msg)
```
jlaasonen: The init method seems redundant.

ebl/tests/transliteration/test_parse_at_line.py
```diff
@@ -73,3 +74,12 @@
 )
 def test_parse_atf_at_line(parser, line, expected_tokens):
     assert parser(line).lines == Text.of_iterable(expected_tokens).lines
+
+
+@pytest.mark.parametrize("parser", [parse_atf_lark])
```
jlaasonen: Why do you parametrize there is only one parser?

jlaasonen:
> > Add test cases for error handling.
> 
> To test the VisitError i would have to do integration testing. Currently only **Duplicate Status** is tested

To test that the `VisitError` is handled properly you would have to update our parser to take the grammar and/or transformer as parameter. Then we could test with a faulty grammar/transformer. This would make the parser more complex and would require dependency injection. Probably too much effort to do it now.

ebl/transliteration/domain/lark_parser.py
```diff
@@ -70,32 +72,8 @@ def parse_line_(line: str, line_number: int):
...
-                None,
-                {"description": f"Invalid line: {ex}", "lineNumber": line_number + 1,},
-            )
-        except EnclosureError:
-            return (
-                None,
-                {"description": f"Invalid brackets.", "lineNumber": line_number + 1,},
-            )
+        except (UnexpectedInput, ParseError, EnclosureError, VisitError) as ex:
+            return catching_lark_error(ex, line, line_number)
```
jlaasonen: The name of the method is misleading. Not all of the exceptions are Lark errors. The method name should tell what it does. Something like `create_transliteration_error_data` would be more informative. You could also just create the data in the method and create the tuple here.

yCobanoglu: Ok the tuple is created in parse_line:
       ` except (UnexpectedInput, ParseError, EnclosureError, VisitError) as ex:
            return (None, create_transliteration_error_data(ex, line, line_number))`

PR #23, opened 2020-03-25
new object structure

yCobanoglu:
fails Travis, this is my idea of implementing the display_value attribute in the API. This would incorporate changing the schema for `DollarLine` and `AtLine`. Alternatively i could calculate the attribute with `@property` from already existing `.content` which could be done without changing the schema.

Also, the `display_token` attribute is similar to the `.atf` of `line` without the `prefix` if im not mistaking.


jlaasonen:
> fails Travis, this is my idea of implementing the display_value attribute in the API. This would incorporate changing the schema for `DollarLine` and `AtLine`. Alternatively i could calculate the attribute with `@property` from already existing `.content` which could be done without changing the schema.
> 
> Also, the `display_token` attribute is similar to the `.atf` of `line` without the `prefix` if im not mistaking.

Looks mostly ok. You could have fixed most of the errors with a little effort and put some more consideration in naming.

It is not possible to do it without changing the schemas, because the display value needs to be serialized. 

It is similar and it's probably to use it to generate the ATF in many cases.

jlaasonen:
[Display mode for Lines beginning with $ and @](https://trello.com/c/IrZ265Ga/388-display-mode-for-lines-beginning-with-and)

ebl/transliteration/application/line_schemas.py
```diff
@@ -31,6 +31,10 @@ def make_line(self, data, **kwargs):
         )
 
 
+class ControlLinesSchema(LineBaseSchema):
```
jlaasonen, line 34: A more descriptive name should be, this has nothing to do with `ControlLine`

ebl/transliteration/application/line_schemas.py
```diff
@@ -31,6 +31,10 @@ def make_line(self, data, **kwargs):
...
+    display_value = fields.String(required=True, data_key="displayValue")
```
jlaasonen, line 35: This cannot be required. It is only needed for dumping and all existing lines will become invalid.

ebl/transliteration/application/at_line_schemas.py
```diff
@@ -53,41 +53,45 @@ def make_label(self, data, **kwargs):
...
 
     @post_load
     def make_line(self, data, **kwargs):
         return ColumnAtLine(data["column_label"])
 
 
-class DiscourseAtLineSchema(LineBaseSchema):
-    discourse_label = NameEnum(atf.Discourse, required=True)
+class DiscourseAtLineSchema(DollarAndAtLineSchema):
+    discourse_label = NameEnum(atf.Discourse, required=True, data_key="discourseLabel")
```
jlaasonen, line 67: You should not change existing property names in schemas without adding fallbacks. Otherwise existing data will become invalid.

ebl/transliteration/application/at_line_schemas.py
```diff
@@ -53,41 +53,45 @@ def make_label(self, data, **kwargs):
...
     def make_line(self, data, **kwargs):
         return DiscourseAtLine(data["discourse_label"])
 
 
-class SurfaceAtLineSchema(LineBaseSchema):
-    surface_label = fields.Nested(SurfaceLabelSchema, required=True)
+class SurfaceAtLineSchema(DollarAndAtLineSchema):
+    surface_label = fields.Nested(
+        SurfaceLabelSchema, required=True, data_key="surfaceLabel"
+    )
```
jlaasonen, line 77: You should not change existing property names in schemas without adding fallbacks. Otherwise existing data will become invalid.

ebl/transliteration/application/at_line_schemas.py
```diff
@@ -53,41 +53,45 @@ def make_label(self, data, **kwargs):
...
     @post_load
     def make_line(self, data, **kwargs):
         return SurfaceAtLine(data["surface_label"])
 
 
-class ObjectAtLineSchema(LineBaseSchema):
+class ObjectAtLineSchema(DollarAndAtLineSchema):
     status = fields.List(NameEnum(atf.Status), required=True)
-    object_label = NameEnum(atf.Object, required=True)
+    object_label = NameEnum(atf.Object, required=True, data_key="objectLabel")
```
jlaasonen, line 86: You should not change existing property names in schemas without adding fallbacks. Otherwise existing data will become invalid.

ebl/transliteration/application/at_line_schemas.py
yCobanoglu: so back to the old names and changing the frontend accordingly ?

jlaasonen: yes

yCobanoglu: it used to extend from LineBaseSchema, i changed it back to that but then i had to add the displayValue attribute to each individual at_line/dollar_line schema

PR #25, opened 2020-04-07
Use clean value when searching lemmas

jlaasonen:


jlaasonen:
[Optimize lemma search](https://trello.com/c/pPG6ClMw/297-optimize-lemma-search)

PR #28, opened 2020-04-23
number_of_lines

yCobanoglu:


ebl/fragmentarium/infrastructure/fragment_repository.py
```diff
@@ -39,11 +39,11 @@ def count_lines(self):
                 {"$unwind": "$text.lines"},
                 {"$replaceRoot": {"newRoot": "$text.lines"}},
                 {"$match": {"type": "TextLine"}},
-                {"$count": "lines"},
+                {"$count": "numbers_of_lines"},
```
jlaasonen, line 42: This does not change anything. See: https://docs.mongodb.com/manual/reference/operator/aggregation/count/

You need to create a new query calculating the sum of text lines from the fragments.

ebl/transliteration/application/text_schema.py
```diff
@@ -7,6 +7,7 @@
 
 class TextSchema(Schema):  # pyre-ignore[11]
     lines = fields.Nested(OneOfLineSchema, many=True, required=True)
+    number_of_lines = fields.Int(required=True)
```
jlaasonen, line 10: The serialized property should be in camel case.

This cannot be required until all the fragments have the field.

ebl/transliteration/domain/text.py
```diff
@@ -32,6 +32,10 @@ class Text:
     lines: Sequence[Line] = tuple()
     parser_version: str = ATF_PARSER_VERSION
 
+    @property
+    def number_of_lines(self) -> int:
+        return len(self.lines)
```
jlaasonen, line 37: We need the number of text lines not the number of all lines. The number must be the same as calculated by the aggregation in `count_lines`.

ebl/transliteration/application/text_schema.py
jlaasonen: This could be manually set zero in the databases prior to deployment and migration, but it needs to be made sure that it does not break the deployed version.

jlaasonen:
[Denormalize statistics](https://trello.com/c/nmyXG6Vc/403-denormalize-statistics)

ebl/fragmentarium/infrastructure/fragment_repository.py
yCobanoglu: Using our new Property 'numberOfLines'  how can i test if its working correct without migrating the DB. Writing a test ?

ebl/transliteration/domain/text.py
yCobanoglu: Currently it's not the same i have to check why.

ebl/transliteration/domain/text.py
```diff
@@ -32,6 +32,14 @@ class Text:
...
     parser_version: str = ATF_PARSER_VERSION
 
+    @property
+    def number_of_lines(self) -> int:
+        counter = 0
+        for line in self.lines:
+            if isinstance(line, TextLine):
+                counter = counter + 1
+        return counter
+
```
jlaasonen, line 42: There is a simpler solution involving list comprehension and `len`.

jlaasonen:
@yCobanoglu what's your plan for migration and deployment? There should be minimal disruption to use of the application.

How long does the full migration take on your computer?

yCobanoglu:
> @yCobanoglu what's your plan for migration and deployment? There should be minimal disruption to use of the application.
> 
> How long does the full migration take on your computer?

25 Minutes. I could use update_fragments with the production database. I could migrate it today at 11 pm for example.

yCobanoglu:
@jlaasonen The json of schemas contains a tuple when dumping (in our tests), then we put the json in the db and if we load it from there it is a list. So currently the json in the tests contain tuples instead list (which would be more accurate). I could convert the tuple into list in the schema should i ?

jlaasonen:
> @jlaasonen The json of schemas contains a tuple when dumping (in our tests), then we put the json in the db and if we load it from there it is a list. So currently the json in the tests contain tuples instead list (which would be more accurate). I could convert the tuple into list in the schema should i ?

JSON cannot contain tuples. The serialized dictionaries should, match that and contain lists instead.

ebl/fragmentarium/update_fragments.py
```diff
@@ -105,10 +105,10 @@ def update_fragments(
     transliteration_factory = context.get_transliteration_update_factory()
     updater = context.get_fragment_updater()
     state = State()
-
+    fragment: Fragment
```
jlaasonen, line 108: This does not work as expected `query_by_fragment_number` fails.  Either an error is raised because `None` does not have `number` or a number of the previous fragment is used.

yCobanoglu: Ok pyre complained about fragments beeing undefined and there is no way of for example declaring it before the for-loop and assigning it None like `fragment = None` do you have a suggesting how to deal with it ?
The error message is (without declaring fragment before the for-loop):
`ebl/fragmentarium/update_fragments.py:115:35 Undefined name [18]: Global name `fragment` is not defined, or there is at least one control flow path that doesn't define `fragment`.`

yCobanoglu: We could pyre ignore it

ebl/transliteration/application/text_schema.py
yCobanoglu: I can set it to `fields.Constant(0)` then merge -> then change the code and migrate and push the commit

ebl/fragmentarium/update_fragments.py
```diff
@@ -107,12 +107,12 @@ def update_fragments(
...
 
     for number in tqdm(numbers, desc=f"Chunk #{id_}", position=id_):
-        fragment = fragment_repository.query_by_fragment_number(number)
         try:
+            fragment = fragment_repository.query_by_fragment_number(number)
             update_fragment(transliteration_factory, updater, fragment)
             state.add_updated()
         except Exception as error:
-            state.add_error(error, fragment)
+            state.add_error(error, fragment)  # pyre-ignore[18]
```
jlaasonen, line 115: The error should be fixed not ignored. The same problem is still there. You just changed the way of ignoring it.

jlaasonen: You could e.g. modify `add_error` to work correctly in all cases, add explicit `expects` for errors requiring the fragment, or add a separate try-catch for querying.

ebl/transliteration/application/text_schema.py
jlaasonen: Migration would make no sense as only fragments with transliteration are updated. The fragments without a transliteration would be left without the property. This is not a problem with your query, but the data quality would be better if all the fragments have the property a correct value. The field can be set with a single query on Mongo shell.

ebl/fragmentarium/update_fragments.py
jlaasonen: We should avoid ignoring errors (see the comment on the new review). Ignores should only be used when Pyre gives an invalid error. E.g.  Pyre does not know the type due to missing type information in a library. Mypy considered these as any-typed but Pyre is stricter.

ebl/fragmentarium/update_fragments.py
yCobanoglu: I added another implementation similar to `add_error` just takes number instead of fragments
```  
def add_querying_error(self, error: Exception, number: str)-> None:
        self.invalid_atf += 1
        self.errors.append(f"{number}\t\t{error}")
```

ebl/transliteration/application/text_schema.py
yCobanoglu: OK  set it to `fields.Constant(0)` -> then merge -> set field with mongo shell -> change number_of_lines` to acutal implementation (get rid of  `fields.Constant(0)`) -> migrate

jlaasonen: There is no need to use `fields.Constant(0)`. Doing merge -> set field -> migrate would have the same effect.

ebl/fragmentarium/update_fragments.py
jlaasonen: You should add a new counter instead. `invalid_atf` which tracks the number of fragments with invalid ATF. If the fragment fails to deserialize it is and another problem.

ebl/transliteration/application/text_schema.py
yCobanoglu: ok and setting the field is done this way ? https://docs.mongodb.com/manual/reference/operator/update/set/

ebl/fragmentarium/update_fragments.py
yCobanoglu: Done. Although there arent really any tests currently that checkmy implementation.

yCobanoglu: i pushed on the wrong branch...

ebl/transliteration/application/text_schema.py
jlaasonen: Yes

PR #37, opened 2020-07-20
validation

yCobanoglu:


ebl/fragmentarium/application/fragment_finder.py
```diff
@@ -41,8 +41,8 @@ def search(self, number: str) -> List[FragmentInfo]:
         )
 
     def inject_documents_in_fragment_infos(
```
jlaasonen, line 43: The name is misleading. The function does the search. Injecting is just an implementation detail.

ebl/fragmentarium/web/fragment_search.py
```diff
@@ -25,9 +22,7 @@ def __init__(
         self._dispatch = create_dispatcher(
             {
                 frozenset(["id", "pages"]): lambda value:
-                finder.inject_documents_in_fragment_infos(
-                    *self._validate_pages(**value)
```
jlaasonen, line 29: The validation was removed. What happens if the input does not represent a number?

yCobanoglu: ok will leave the validation i though validating on the frontend is enough

jlaasonen: Never trust data from the web.

PR #39, opened 2020-07-29
Use clean values to generate signs

jlaasonen:


jlaasonen:
[Use clean values in sign mapping](https://trello.com/c/4bvY7zmj/338-use-clean-values-in-sign-mapping)

PR #43, opened 2020-08-21
Select genre

yCobanoglu:


jlaasonen:
[Select Genre in Fragmentarium](https://trello.com/c/ps6PD12L/331-select-genre-in-fragmentarium)

ebl/fragmentarium/web/fragment_genre.py
```diff
@@ -0,0 +1,21 @@
+import falcon  # pyre-ignore
```
jlaasonen, line 1: Please ignore a specific error.

jlaasonen (changes requested):
You say that the genre is a tree structure, but here it is a nested list of strings. There should be validation that the genres are correct. Now it would be possible to set e.g. to [["foo", "bar"]]. 

If the genre is defined by a tree, saving  the parent information in the fragment seems redundant. Also, changing the tree would require updating all the fragments.

yCobanoglu:
> You say that the genre is a tree structure, but here it is a nested sequence of strings. There should be validation that the genres are correct. Now it would be possible to set e.g. to `[["foo", "bar"]]`.
> 
> If the genre is defined by a tree, saving the parent information in the fragment seems redundant. Also, changing the tree would require updating all the fragments.

@jlaasonen 
I can make a validation. But saving only the **leaf** of the trees in the database, how could one search for the **nodes** of that **leaf** in the database ?

jlaasonen:
> > You say that the genre is a tree structure, but here it is a nested sequence of strings. There should be validation that the genres are correct. Now it would be possible to set e.g. to `[["foo", "bar"]]`.
> > If the genre is defined by a tree, saving the parent information in the fragment seems redundant. Also, changing the tree would require updating all the fragments.
> 
> @jlaasonen
> I can make a validation. But saving only the **leaf** of the trees in the database, how could one search for the **nodes** of that **leaf** in the database ?

Is there a requirement to search **leaves** or **nodes**? What we have now is just a bunch of strings there is are no **leaves** and **nodes**. If we need to search **genres** the query can be constructed from the genre configuration. You can keep the denormalized format if you think it is better, but you must ensure that it is correct all the time.

jlaasonen:
About modelling trees in mongo: https://docs.mongodb.com/manual/applications/data-models-tree-structures/

ebl/fragmentarium/web/bootstrap.py
```diff
@@ -58,13 +62,15 @@ def create_fragmentarium_routes(api: falcon.API, context: Context, spec):  # pyr
...
     api.add_route("/fragments", fragment_search)
+    api.add_route("/fragments/{number}/genre", fragment_genre)
     api.add_route("/fragments/{number}", fragments)
     api.add_route("/fragments/{number}/pager", fragment_pager)
     api.add_route("/fragments/{number}/lemmatization", lemmatization)
     api.add_route("/fragments/{number}/references", references)
     api.add_route("/fragments/{number}/transliteration", transliteration)
     api.add_route("/fragments/{number}/annotations", annotations)
     api.add_route("/fragments/{number}/photo", photo)
+    api.add_route("/genre", genre)
```
jlaasonen, line 73: Resource name should `genres`.

ebl/fragmentarium/web/fragment_genre.py
```diff
@@ -0,0 +1,34 @@
...
+from ebl.users.web.require_scope import require_scope
+
+
+class FragmentGenreResource:
+    def __init__(self, updater: FragmentUpdater):
+        self._updater = updater
+
+    @staticmethod
+    def _validate_genre(genres_retrieved: Sequence[Sequence[str]]):
+        return all(genre in genres for genre in genres_retrieved)
```
jlaasonen, line 17: The validation should be in the fragment model.

ebl/fragmentarium/web/fragment_genre.py
```diff
@@ -0,0 +1,34 @@
...
+                parse_museum_number(number),
+                req.media["genre"], user
+            )
+            resp.media = create_response_dto(updated_fragment, user, has_photo)
+        else:
+            resp.status = falcon.HTTP_UNPROCESSABLE_ENTITY
+            resp.media = {
+                "title": resp.status,
+                "description": "Not a valid Genre",
+            }
```
jlaasonen, line 34: If you raise a `DataError` the error handler will take care of the response.

ebl/tests/fragmentarium/test_fragment.py
```diff
@@ -144,6 +144,18 @@ def test_references_default():
...
+def test_genre():
+    genre = (("ARCHIVE", "Administrative", "Lists", "One Entry"),)
+    fragment = FragmentFactory.build(genre=genre)
+    assert fragment.genre == genre
+
+
+def test_genre_empty():
+    genre = tuple()
+    fragment = FragmentFactory.build(genre=genre)
+    assert fragment.genre == genre
```
jlaasonen, line 156: This is redundant. Getting the genre is already tested above.

ebl/tests/fragmentarium/test_fragment.py
```diff
@@ -217,3 +229,10 @@ def test_set_references():
     updated_fragment = fragment.set_references(references)
 
     assert updated_fragment.references == references
+
+
+def test_set_genre():
```
jlaasonen, line 234: Put this together with the other genre tests.

ebl/tests/fragmentarium/test_fragment_repository.py
```diff
@@ -156,6 +157,18 @@ def test_update_update_transliteration_not_found(fragment_repository):
         fragment_repository.update_transliteration(transliterated_fragment)
 
 
+def test_update_genre(fragment_repository):
+    fragment = FragmentFactory.build(genre=tuple())
+    fragment_number = fragment_repository.create(fragment)
+    updated_fragment = fragment.set_genre([["ARCHIVE", "Administrative"]])
+    fragment_repository.update_genre(updated_fragment)
+    result = fragment_repository.query_by_museum_number(
+        parse_museum_number(fragment_number)
```
jlaasonen, line 166: Use `fragment.number` instead of parsing the `_id`.

ebl/fragmentarium/domain/genre.py
```diff
@@ -0,0 +1,6 @@
+genres = [
```
jlaasonen, line 1: Shouldn't there be more genres?

jlaasonen (changes requested):

ebl/fragmentarium/domain/fragment.py
```diff
@@ -65,6 +68,9 @@ def update_transliteration(
             record=record,
         )
 
+    def set_genre(self, genre: Genre) -> "Fragment":
+        return attr.evolve(self, genre=tuple(map(tuple, genre)))  # pyre-ignore[6]
```
jlaasonen, line 72: Pyre errors should not be arbitrarily ignored. In general, only errors from libraries without types should be ignored.

Mypy had an option to treat them as `Any`, but pyre likes to complain about them whenever possible.

ebl/fragmentarium/domain/genre.py
jlaasonen: Genres should be `tuple`s to avoid accidental modification.

jlaasonen (changes requested):

ebl/fragmentarium/domain/fragment.py
yCobanoglu: i cercumvented this error py creating a new function and adding this link as a comment.
[https://github.com/python/mypy/issues/1317](url)
weirdly enough `list(map(list, genre))` works (that is used in **test_genre_route.py**.

ebl/fragmentarium/domain/fragment.py
```diff
@@ -1,16 +1,25 @@
...
 from ebl.fragmentarium.domain.transliteration_update import TransliterationUpdate
 from ebl.transliteration.domain.lemmatization import Lemmatization
 from ebl.transliteration.domain.text import Text
 from ebl.users.domain.user import User
 from ebl.fragmentarium.domain.museum_number import MuseumNumber
 
+Genre = Sequence[Sequence[str]]
+
+
+# https://github.com/python/mypy/issues/1317
```
jlaasonen, line 19: Mypy does not work with Pypy. Does this also affect Pyre?

ebl/fragmentarium/domain/fragment.py
```diff
@@ -65,6 +75,20 @@ def update_transliteration(
...
+            return True
+        else:
+            return False
+
+    def set_genre(self, genre_retrieved: Genre) -> "Fragment":
+        genre_retrieved = tuple(map(convert_genre_elem_to_tuple, genre_retrieved))
+        if Fragment._is_genre_valid(genre_retrieved):
+            return attr.evolve(self, genre=genre_retrieved)
+        else:
+            raise ValueError(f"'{(genre_retrieved)}' is not a valid genre")
```
jlaasonen, line 90: You could use [validator](https://www.attrs.org/en/stable/examples.html?highlight=validator#validators). Now it is possible to set incorrect genres during init.

jlaasonen (changes requested):

ebl/fragmentarium/domain/fragment.py
yCobanoglu: i changed it to **List comprehenion**. I couldnt find anything concerning this issue for **pyre**.
Also we have the same code (namely: `tuple(map(tuple, ...)`in **fragment_info_schema.py** and a similar line with `list`instead of `tuple`in **test_genre_route.py**. No clue whats the issue with **pyre**.

ebl/fragmentarium/domain/fragment.py
yCobanoglu: oh yeah could call would have totally missed that **!**

yCobanoglu: So whats weird is that the validator takes a genre argument in initialzation which is `Tuple[Tuple[str,...],...]` but you can pass it a `list` and it wont throw a `TypeError`.

ebl/fragmentarium/domain/fragment.py
```diff
@@ -65,6 +75,13 @@ def update_transliteration(
             record=record,
         )
 
+    def set_genre(self, genre_retrieved: Sequence[Sequence[str]]) -> "Fragment":
+        genre_retrieved = tuple(
+            [tuple(single_genre) for single_genre in genre_retrieved]
+        )
+        self._check_is_genre_valid(None, genre_retrieved)
```
jlaasonen, line 82: This should not be necessary. See: https://www.attrs.org/en/stable/api.html?highlight=evolve#attr.evolve

jlaasonen (approved):

ebl/fragmentarium/domain/fragment.py
yCobanoglu: ok

yCobanoglu:
unfortunatley i got the genre "data" model wrong every genre needs a property uncertain which can be true or false. After reading the trello card today it became clear to me it was my fault but anyways the backend is ready a genre is now a class.

ebl/fragmentarium/domain/fragment.py
jlaasonen: Types are not checked runtime.

jlaasonen (approved):

PR #45, opened 2020-09-04
Parse museum number

jlaasonen:


jlaasonen:
[Parse museum number](https://trello.com/c/63iPDfSh/396-parse-museum-number)

PR #48, opened 2020-09-16
Use line number model in reconstructed line

jlaasonen:


jlaasonen:
[Parse reconstructed line number](https://trello.com/c/S4jDltIH/452-parse-reconstructed-line-number)

PR #53, opened 2020-10-05
mongosync script

yCobanoglu:


yCobanoglu:
@jlaasonen everything is working **besides** environment variables. They aren't passed to the container/ the script. If i hardcode the env variables in the mongosync.sh  it works but if i try to access then through docker-compose it fails.

mongosync/mongosync.sh
```diff
@@ -0,0 +1,51 @@
+#!/bin/bash
+
+printenv
```
jlaasonen, line 3: The environment contains secrets. Do not print it.

jlaasonen (changes requested):
You removed the old script, but the new one does not have all the feature. E.g. allow any DB or use different host and user for source and target.

Either implement the missing features and update the README or move the script to own repository.

jlaasonen:
[Synching Ebl-dev database with bash script](https://trello.com/c/eUptlA9M/470-synching-ebl-dev-database-with-bash-script)

jlaasonen:
Moved to own repository: https://github.com/ElectronicBabylonianLiterature/mongosync

PR #86, opened 2021-02-01
Indication of parallel lines

jlaasonen:


jlaasonen:
[Indication of parallel lines](https://trello.com/c/CAkQtcMX/288-indication-of-parallel-lines)

docs/ebl-atf.md
```diff
@@ -161,6 +161,18 @@ note-text = { note-character };
...
+## Parallel lines
+
+```ebnf
+parallel-line = '// ', [ 'cf. ' ], ( parallel-composition | parallel-text | parallel-fragment );
+parallel-composition = '(',  { any-character }-, ' ', line-number,  ')';
+parallel-text = 'C ', category, '.', index, ' ', [ stage, ' ', { any-character }-, ' ', ], line-number;
+parallel-fragment = [ '&d ' ], museum-number, ' ', [ surface-label, ' ' ], line-number;
+
+museum-number = ? .+?\.[^.]+(\.[^.]+)? ?;
+stage = 'Ur3' | 'OA' | 'OB' | 'MB' | 'MA | 'Hit | 'NA' | 'NB' | 'LB' | 'Per' | 'Hel' | 'Par' | 'Unc' | 'SB';
```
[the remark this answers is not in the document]
ejimsan: [ '&d ' ] should go after the `museum-number`

docs/ebl-atf.md
```diff
@@ -161,6 +161,18 @@ note-text = { note-character };
 note-character = any-character - ( '@' | '{' | '}' );
 ```
 
+## Parallel lines
+
+```ebnf
+parallel-line = '// ', [ 'cf. ' ], ( parallel-composition | parallel-text | parallel-fragment );
+parallel-composition = '(',  { any-character }-, ' ', line-number,  ')';
+parallel-text = 'C ', category, '.', index, ' ', [ stage, ' ', { any-character }-, ' ', ], line-number;
```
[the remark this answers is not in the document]
ejimsan: Instead of `'C '`, it might be more meaningful to use `L ` (= "Literature"). When we adapt the Corpus for other genres, then we can introduce one letter for each corpus.

docs/ebl-atf.md
acheinrich, line 174: It should be `'MA'` and `'Hit` if I am not mistaken.

jlaasonen:
[Indication of parallel lines](https://trello.com/c/CAkQtcMX/288-indication-of-parallel-lines)

docs/ebl-atf.md
```diff
@@ -161,6 +161,23 @@ note-text = { note-character };
...
 ```
 
+## Parallel lines
+
+```ebnf
+parallel-line = '// ', [ 'cf. ' ],
+                ( parallel-composition | parallel-text | parallel-fragment );
+parallel-composition = '(',  { any-character }-, ' ', line-number,  ')';
+parallel-text = corpus-type, ' ', category, '.', index, ' ',
+                [ stage, ' ', { any-character }-, ' ', ], line-number;
```
acheinrich, line 171: Would it be easier to understand if we put `[stage, ' ', version,  ' ', ], line number;` here and then define `version = { any character };` after `stage = 'Ur3'  | ...;` ?

docs/ebl-atf.md
```diff
@@ -161,6 +161,23 @@ note-text = { note-character };
...
+parallel-fragment = 'F ', museum-number, [ '&d ' ], ' ', [ surface-label, ' ' ],
+                    line-number;
+
+museum-number = ? .+?\.[^.]+(\.[^.]+)? ?;
+corpus-type = 'L'
```
acheinrich, line 176: Maybe change the name from `corpus-type` to `genre`, which might be easier to grasp for assyriologists? What do you think @ejimsan ?

ejimsan: Yes, that's a good idea

docs/ebl-atf.md
jlaasonen: It does not make sense to have space after an empty version and name should also be there. E.g:

```
parallel-text = `genre, ' ', category, '.', index, ' ',  [ stage, ' ',  [ version, ' ' ], name, ' '], line-number
version  = { any-character }-;
name =  { any-character }-;
```

But since version and name can be any string the syntax would be ambiguous. Also, it would not be possible to distinguish empty and omitted version without adding more syntax. E.g.: `version  = '"' { any-character } '"';`.

acheinrich: 1)  It is an excellent idea to put the content of `version` and `name` in `"` as the semantic implications of `"` would point a reader into the right direction and show that these are conventionally chosen designation not only on the level of technical definition but also editorially.
2) `name` should be part of the definition of `parallel-composition` and not of `parallel-text`, or am I not getting something?

jlaasonen: image Chapters have name. It is called`Chapter_Number` in the conventions, but does not have to a number.

acheinrich: All right, now I get it. Thank you for helping me understand. So, if chapters are always referenced by their name maybe we should change `name` to `chapter`?

jlaasonen: Would the grammar be then:

```
parallel-text = `genre, ' ', category, '.', index, ' ',  [ stage, ' ',  [ version, ' ' ], chapter , ' ' ], line-number
version  = '"', { any-character }-, '"';
chapter =  '"', { any-character }-, '"';
```

There would still be some ambiguity, but it is less likely to cause problems. It could be removed by escaping `"` if it appears in version or chapter.

acheinrich: Looks good to me.

PR #99, opened 2021-03-18
Strip alignments only if alignable indices change

jlaasonen:


jlaasonen:
[Update check if reconstruction has changed.](https://trello.com/c/19GwSezv/513-update-check-if-reconstruction-has-changed)

PR #100, opened 2021-03-21
added mesZl and logogram property to Sign Class and Sign Schema

yCobanoglu:


ebl/transliteration/infrastructure/mongo_sign_repository.py
```diff
@@ -34,15 +39,31 @@ def filter_none(self, data, **kwargs):
...
+        data["word_id"] = tuple(data["word_id"])
+        return Logogram(**data)
+
+
 class SignSchema(Schema):
     name = fields.String(required=True, data_key="_id")
     lists = fields.Nested(SignListRecordSchema, many=True, required=True)
     values = fields.Nested(ValueSchema, many=True, required=True, unknown=EXCLUDE)
+    logograms = fields.Nested(LogogramSchema, many=True)
+    mes_zl = fields.String(data_key="mesZl")
```
jlaasonen, line 59: It would be clearer and less error-prone to use `missing` instead of relying on the default value in the `Sign` constructor.

ebl/transliteration/infrastructure/mongo_sign_repository.py
```diff
@@ -34,15 +39,31 @@ def filter_none(self, data, **kwargs):
...
 
     @post_load
     def make_sign(self, data, **kwargs) -> Sign:
         data["lists"] = tuple(data["lists"])
         data["values"] = tuple(data["values"])
+        if data.get("logograms") is not None:
```
jlaasonen, line 65: It would be simpler to use `missing` for the field.

jlaasonen (changes requested):

jlaasonen (approved):

PR #101, opened 2021-03-22
Preserve alignments when changing reconstruction

jlaasonen:


jlaasonen:
[Preserve alignments when changing reconstructed line](https://trello.com/c/kBS52Hdv/509-preserve-alignments-when-changing-reconstructed-line)

PR #107, opened 2021-04-15
add seal to common in ebl-atf.md

yCobanoglu:


jlaasonen (changes requested):
Add also the seal dollar-line.

jlaasonen (approved):

PR #120, opened 2021-06-09
Adapt Corpus for other genres

jlaasonen:


jlaasonen:
[Adapt Corpus for other Genres](https://trello.com/c/PmR2pD6e/497-adapt-corpus-for-other-genres)

PR #121, opened 2021-06-10
find sign by signName endpoint

yCobanoglu:


yCobanoglu:
[Create interface for signs](https://trello.com/c/Ss2PYZ15/492-create-interface-for-signs)

ebl/signs/web/bootstrap.py
```diff
@@ -2,8 +2,11 @@
...
 from ebl.context import Context
 from ebl.signs.web.sign_search import SignsSearch
+from ebl.signs.web.signs import SignsResource
 
 
 def create_signs_routes(api: falcon.API, context: Context):
     signs_search = SignsSearch(context.sign_repository)
+    signs = SignsResource(context.sign_repository)
     api.add_route("/signs", signs_search)
+    api.add_route("/signs/{object_id}", signs)
```
jlaasonen, line 12: Why is the parameter called "object_id" when it is not `ObjectId`?

jlaasonen (changes requested):

ebl/signs/web/bootstrap.py
yCobanoglu: i used it so it is similar to the words route. I will change it.

jlaasonen (approved):

ebl/signs/web/bootstrap.py
jlaasonen: Words route is the first route implemented. The name is probably a leftover from a time when `ObjectId` was used and should be updated.

PR #128, opened 2021-08-04
Add an endpoint for retrieving colophons

jlaasonen:


jlaasonen:
[Implement display for frontpage of text in Corpus](https://trello.com/c/9xzVmoaq/529-implement-display-for-frontpage-of-text-in-corpus)

PR #138, opened 2021-08-26
Infer dispatch parameters from command signatures

jlaasonen:


jlaasonen:
@yCobanoglu @khoidt @fsimonjetz 

Do you think it makes sense to change the dispatcher to work like this? If not, please close this PR.

PR #143, opened 2021-09-10
Use pydash to group extant lines

jlaasonen:
itertools.groupby requires that the input is soretd.


jlaasonen:
[Implement display for frontpage of text in Corpus](https://trello.com/c/9xzVmoaq/529-implement-display-for-frontpage-of-text-in-corpus)

PR #145, opened 2021-09-27
Incorporate AGI

ejimsan:


ebl/dictionary/application/word_schema.py
```diff
@@ -67,6 +67,14 @@ class OraccWordSchema(Schema):
     guideWord = fields.String(required=True)
 
 
+class akkadischeGlossareUndIndices(Schema):
```
jlaasonen, line 70: Use [PEP8 naming conventions](https://www.python.org/dev/peps/pep-0008/#class-names). I.e. CapWords for class names.

Schema class names should end with Schema for consistency and to avoid conflicts with model classes.

jlaasonen (changes requested):

jlaasonen (approved):

ebl/dictionary/application/word_schema.py
ejimsan: Done, thank you.

PR #147, opened 2021-10-01
Allow translations in Corpus import

jlaasonen:


jlaasonen:
[Add translations to corpus importer](https://trello.com/c/l83w853d/542-add-translations-to-corpus-importer)

PR #160, opened 2021-11-03
add enum

yCobanoglu:


ebl/fragmentarium/domain/annotation.py
```diff
@@ -22,6 +22,7 @@ class AnnotationValueType(Enum):
     RULINGDOLLARLINE = "RulingDollarLine"
     BLANK = "Blank"
     PREDICTED = "Predicted"
+    PARTIALLYBROKEN = "PartiallyBroken"
```
jlaasonen, line 25: Separate words with an underscore.

jlaasonen (changes requested):

ebl/fragmentarium/domain/annotation.py
jlaasonen: https://www.python.org/dev/peps/pep-0008/#constants

yCobanoglu: ok.

jlaasonen (approved):

PR #164, opened 2021-11-08
save files with bounding boxes

yCobanoglu:


ebl/fragmentarium/retrieve_annotations.py
```diff
@@ -3,13 +3,52 @@
...
+
+
+def filter_annotation(annotation: Annotation) -> bool:
+    if (
+        annotation.data.type != AnnotationValueType.RULINGDOLLARLINE
+        and annotation.data.type != AnnotationValueType.SURFACEATLINE
+    ):
+        return True
+    else:
+        return False
```
jlaasonen, line 28: Returning `True` or `False` is redundant. Just return the value of the condition. See also: https://github.com/ElectronicBabylonianLiterature/ebl-api/pull/165

jlaasonen (changes requested):

jlaasonen (approved):

PR #170, opened 2021-11-29
Add endpoint for chapter display

jlaasonen:


jlaasonen:
[Implement basic display for corpus text](https://trello.com/c/ZtKcwlnJ/544-implement-basic-display-for-corpus-text)

PR #172, opened 2021-11-30
Signs images

yCobanoglu:


yCobanoglu:
[Display annotations in Signs interface](https://trello.com/c/C3BG2qRF/538-display-annotations-in-signs-interface)

ebl/fragmentarium/application/annotations_image_extractor.py
```diff
@@ -0,0 +1,126 @@
...
+        if isinstance(line, TextLine):
+            return label, [*lines, [label.set_line_number(line.line_number), line]]
+        elif isinstance(line, ObjectAtLine):
+            return label.set_object(line.label), lines
+        elif isinstance(line, SurfaceAtLine):
+            return label.set_surface(line.surface_label), lines
+        elif isinstance(line, ColumnAtLine):
+            return label.set_column(line.column_label), lines
+        else:
+            return label, lines
```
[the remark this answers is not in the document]
jlaasonen: This should be fixed. You could use e.g. single dispatch.

ebl/fragmentarium/application/annotations_image_extractor.py
```diff
@@ -0,0 +1,126 @@
...
+        fragment_repository: FragmentRepository,
+        annotations_repository: AnnotationsRepository,
+        photos_repository: FileRepository,
+    ):
+        self._fragments_repository = fragment_repository
+        self._annotations_repository = annotations_repository
+        self._photos_repository = photos_repository
+
+    @staticmethod
+    def _calculate_label(total: Tuple[LineLabel, Sequence[Line]], line: Line):
```
jlaasonen, line 43: The return type is missing.

ebl/fragmentarium/application/annotations_image_extractor.py
```diff
@@ -0,0 +1,126 @@
...
+            bounding_boxes.top_left_y + bounding_boxes.height,
+        )
+        cropped_image = image.crop(area)
+        buf = io.BytesIO()
+        cropped_image.save(buf, format="PNG")
+        return Base64(base64.b64encode(buf.getvalue()).decode("utf-8"))
+
+    def _cropped_image_from_annotations(
+        self, fragment_number: MuseumNumber, annotations: Sequence[Annotation]
+    ):
```
jlaasonen, line 98: The return type is missing.

ebl/tests/conftest.py
```diff
@@ -196,7 +196,7 @@ def close(self) -> None:
         self._file.close()
 
     def read(self, size=-1):
-        return self._file.read(size)
+        return self._file.getvalue()
```
jlaasonen, line 199: Does this work if the size is not -1?

ebl/tests/fragmentarium/test_annotations_image_extractor.py
```diff
@@ -0,0 +1,45 @@
...
+        .query_by_file_name(f"{annotation.fragment_number}.jpg")
+        .thenReturn(photo_jpeg)
+    )
+
+    result = image_extractor.cropped_images_from_sign(sign)
+    assert len(result) > 0
+    first_cropped_annotation = result[0]
+    assert isinstance(first_cropped_annotation, CroppedAnnotation)
+    assert first_cropped_annotation.script == fragment.script
+    assert first_cropped_annotation.label != "i stone wig i 8'"
```
jlaasonen, line 45: What is the purpose of this assertion?
Wouldnt it be better to test that the label is correct?

ebl/tests/fragmentarium/test_annotations_repository.py
```diff
@@ -6,6 +6,19 @@
...
+    annotations = AnnotationsFactory.build_batch(5)
+    sign_query = annotations[0].annotations[0].data.sign_name
+    database[COLLECTION].insert_many(AnnotationsSchema(many=True).dump(annotations))
+
+    results = annotations_repository.find_by_sign(sign_query)
+
+    assert len(results) >= 1
+    for result in results:
+        for annotation in result.annotations:
+            assert annotation.data.sign_name == sign_query
```
jlaasonen, line 19: The test is incomplete. It would not catch e.g. spurious annotations or incorrect values of other properties.

ebl/tests/signs/test_sign_images_route.py
```diff
@@ -0,0 +1,37 @@
...
+
+    result = client.simulate_get(f"/signs/{sign_name}/images")
+
+    assert len(result.json) > 0
+    result_json = result.json[0]
+
+    assert result_json["fragmentNumber"] == str(fragment.number)
+    assert isinstance(result_json["image"], str)
+    assert result_json["script"] == fragment.script
+    assert isinstance(result_json["label"], str)
```
jlaasonen, line 34: Why do you only test the result partially?

ebl/tests/transliteration/test_line_label.py
```diff
@@ -0,0 +1,79 @@
...
+        ),
+        (
+            LineLabel(
+                None, None, None, LineNumberRange(LineNumber(1, True), LineNumber(3))
+            ),
+            "1'-3",
+        ),
+    ],
+)
+def test_line_label(line_label, expected):
```
jlaasonen, line 29: The test name is misleading. You are only testing abbreviation property.

ebl/transliteration/domain/line_label.py
```diff
@@ -24,3 +28,38 @@ def set_object(self, object: Optional[ObjectLabel]) -> "LineLabel":
 
     def set_line_number(self, line_number: Optional[AbstractLineNumber]) -> "LineLabel":
         return attr.evolve(self, line_number=line_number)
+
+    def is_line_number(self, line_number_to_match: int) -> bool:
```
jlaasonen, line 32: The name is confusing as it does not check if something is a line number. `has_line_number` might be better.

What is the purpose of this method? It ignores prime, prefix and suffix and an index from the annotation path is passed in.  https://github.com/ElectronicBabylonianLiterature/ebl-api/blob/b6889aaa72f47eb7216e8f5649f487792674445d/ebl/fragmentarium/application/annotations_image_extractor.py#L96)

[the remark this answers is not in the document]
jlaasonen: This should be fixed.

jlaasonen (changes requested):

ebl/fragmentarium/application/annotations_image_extractor.py
jlaasonen: The missing type hides problems with the implementation. The method is used with `reduce` which means that the type of the accumulator and return value should be the same. That is not the case in line 47 which returns `Tuple[LineLabel, Sequence[Union[Line, List[Union[LineLabel, Line]]]`.

jlaasonen: The logic is mostly duplicated from https://github.com/ElectronicBabylonianLiterature/ebl-api/blob/bf802db8b789979875856abafdd950ae41b93c5b/ebl/transliteration/domain/text.py#L105

The code should be changed so that there is no duplicate logic.

ebl/fragmentarium/application/annotations_image_extractor.py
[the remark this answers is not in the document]
yCobanoglu: ok yeah i copied this one to one from javascript i will do single dispatch. Single dispatch needs so much code in python.

ebl/tests/conftest.py
yCobanoglu: No so i one wants to only read specific number of bytes it won't work. The problem previously was that once read the buffer doesn't reset itself so for proceeding tests `read` returned nothing.  What should i do ?

ebl/tests/fragmentarium/test_annotations_repository.py
yCobanoglu: but why would i test anything beyond the sign_name if something was incorrect and should have been prevented way earlier than here or am i missing something ? The method should retrieve every annotation that has that specific sign_name only.

ebl/tests/conftest.py
jlaasonen: Do not use the same file in multiple tests, if they change its state.

jlaasonen: If you return the whole file everytime, callers reading the file in chunks may not work as intended.

jlaasonen (changes requested):

jlaasonen (approved):

PR #177, opened 2021-12-17
Switch from pipenv to poetry

jlaasonen:
https://python-poetry.org/


jlaasonen:
Poetry does not and will not support running scripts like pipenv or npm. It is possible by using a plugin (e.g. Poe) or external tool (e.g. Make). It is probably better to use a tool that has this feature, e.g. PDM.

PR #184, opened 2022-01-20
Update to PyPy 3.8

jlaasonen:


jlaasonen:
Joblib currently does not work on Python 3.8: https://github.com/joblib/joblib/issues/1238

PR #232, opened 2022-04-20
Old siglum

fsimonjetz:


ebl/corpus/application/schemas.py
```diff
@@ -52,9 +53,25 @@
...
+    @post_load
+    def make_old_siglum(self, data, **kwargs) -> OldSiglum:
+        return OldSiglum(**data)
+
+
 class ManuscriptSchema(Schema):
     id = fields.Integer(required=True)
     siglum_disambiguator = fields.String(required=True, data_key="siglumDisambiguator")
+    old_siglum = fields.Nested(
+        OldSiglumSchema,
```
yCobanoglu, line 69: If Manuscript can have multiple `old_siglum` maybe naming it `old_siglums` instead ?

ebl/corpus/web/chapter_schemas.py
```diff
@@ -61,7 +62,18 @@ def _deserialize_transliteration(value):
         raise ValidationError(f"Invalid colophon: {value}.", "colophon") from error
 
 
+class ApiOldSiglumSchema(OldSiglumSchema):
+    reference = fields.Nested(ApiReferenceSchema, required=True)
+
```
yCobanoglu, line 67: `Can we name it only OldSiglumSchema` without the Api ?

ebl/corpus/web/chapter_schemas.py
```diff
@@ -61,7 +62,18 @@ def _deserialize_transliteration(value):
...
+
 class ApiManuscriptSchema(ManuscriptSchema):
+    old_siglum = fields.Nested(
```
yCobanoglu, line 70: same thing here.

ebl/tests/corpus/test_api_chapter_schema.py
```diff
@@ -133,6 +137,7 @@ def test_serialize_manuscript() -> None:
     assert ApiManuscriptSchema().dump(manuscript) == {
         "id": manuscript.id,
         "siglumDisambiguator": manuscript.siglum_disambiguator,
+        "oldSiglum": ApiOldSiglumSchema().dump(manuscript.old_siglum, many=True),
         "museumNumber": str(manuscript.museum_number)
```
yCobanoglu, line 141: I think would be better to have
```python
"oldSiglum": {...} 
```
That way the test is not reliying on `ApiOldSiglumSchema` to be correct.

yCobanoglu, line 141: There should be a test for `ApiOldSiglumSchema()` or you could write `oldSiglum: { ... }`

ebl/tests/corpus/test_api_chapter_schema.py
```diff
@@ -133,6 +137,7 @@ def test_serialize_manuscript() -> None:
     assert ApiManuscriptSchema().dump(manuscript) == {
         "id": manuscript.id,
         "siglumDisambiguator": manuscript.siglum_disambiguator,
+        "oldSiglum": ApiOldSiglumSchema().dump(manuscript.old_siglum, many=True),
```
yCobanoglu, line 140: There should be a test for `ApiOldSiglumSchema()` or you could write `oldSiglum: { ... }`

ebl/corpus/domain/manuscript.py
```diff
@@ -118,6 +118,12 @@ def __str__(self) -> str:
         )
 
 
+@attr.s(auto_attribs=True, frozen=True)
+class OldSiglum:
+    siglum: str = ""
+    reference: Optional[Reference] = None
+
```
yCobanoglu, line 125: Remove default values

ebl/corpus/domain/manuscript.py
```diff
@@ -134,6 +140,7 @@ def is_invalid_non_standard_text(provenance, period, type_) -> bool:
 class Manuscript:
     id: int
     siglum_disambiguator: str = ""
+    old_siglum: Sequence[OldSiglum] = tuple()
     museum_number: Optional[MuseumNumber] = None
```
yCobanoglu, line 144: Make it plural

ebl/tests/corpus/test_api_chapter_schema.py
```diff
@@ -151,13 +156,16 @@ def test_serialize_manuscript() -> None:
 
 
 def test_deserialize_manuscript() -> None:
-    references = (ReferenceFactory.build(with_document=False),)
+    references = (ReferenceFactory.build(with_document=True),)
     manuscript = ManuscriptFactory.build(references=references)
```
yCobanoglu, line 160: Maybe leave it `False` unless there is a reason

ebl/tests/corpus/test_chapter_schema.py
```diff
@@ -93,6 +94,9 @@ def to_dict(chapter: Chapter, include_documents=False):
             {
                 "id": manuscript.id,
                 "siglumDisambiguator": manuscript.siglum_disambiguator,
+                "oldSiglum": (
+                    ApiOldSiglumSchema if include_documents else OldSiglumSchema
+                )().dump(manuscript.old_siglum, many=True),
                 "museumNumber": (
```
yCobanoglu, line 100: extract variable.

ebl/tests/factories/corpus.py
```diff
@@ -48,12 +49,21 @@
...
+
+
 class ManuscriptFactory(factory.Factory):
     class Meta:
         model = Manuscript
 
     id = factory.Sequence(lambda n: n + 1)
     siglum_disambiguator = factory.Faker("word")
+    old_siglum = factory.List([factory.SubFactory(OldSiglumFactory)], TupleFactory)
     museum_number = factory.Sequence(
```
yCobanoglu, line 67: @jlaasonen  how about `tuple(factory.List(...)` ?

jlaasonen: Probably does not work, but you can try.

fsimonjetz: According to the  [docs](https://factoryboy.readthedocs.io/en/stable/reference.html?highlight=factory.List#factory.List), `factory.List` is used for arbitrary "list-like attributes" (lists, tuples, sets, ...). The second parameter `list_factory` determines the type.

ebl/tests/corpus/test_api_chapter_schema.py
fsimonjetz: The purpose of `ApiOldSiglumSchema` is to use `ApiReferenceSchema` for the `reference` field, while the regular `OldSiglumSchema` uses the regular `ReferenceSchema`. However, the only difference between `ApiReferenceSchema` and `ReferenceSchema` seems to be that the `document` field is set to `load_only=False` (the default) for the former, while for the latter it's set to `load_only=True`. 
IIUC, we could get rid of one of the two schemas and explicitly exclude `document` at the appropriate places, either by setting `exclude=('document',)` or `load_only=('document',)` when *instantiating* the schema for serializing, instead of in the schema definition. It would require some refactoring, but I think then we could get rid of `ApiOldSiglumSchema`, too.

fsimonjetz:
@yCobanoglu I'd be grateful if you could take another look. Some more extensive changes were necessary for implementing the `reference` embedded in the `OldSiglum` instances, testing the schemas and passing additional manuscript attributes to the line display in the frontend.

jlaasonen (approved):

PR #235, opened 2022-04-21
Optimize corpus search

khoidt:


khoidt:
[Optimize Corpus search](https://trello.com/c/ATShlUQe/564-optimize-corpus-search)

PR #274, opened 2022-07-14
Add old line numbers

fsimonjetz:
Implements https://trello.com/c/kfscPvSj


ebl/tests/corpus/test_chapter.py
```diff
@@ -69,6 +69,7 @@
 )
 REFERENCES = (ReferenceFactory.build(),)
 LINE_NUMBER = LineNumber(1)
+OLD_LINE_NUMBERS = tuple()
```
khoidt, line 72: Perhaps would make sense to test this with data.

ebl/tests/corpus/test_chapter_merge.py
```diff
@@ -66,9 +66,11 @@ def test_merge_manuscript_line(old, new, expected):
 IS_SECOND_LINE_OF_PARALLELISM = True
 IS_BEGINNING_OF_SECTION = True
 NOTE = None
+OLD_LINE_NUMBERS = tuple()
```
khoidt, line 69: Perhaps would make sense to test this with data.

ebl/tests/corpus/test_chapter_display.py
```diff
@@ -19,6 +19,7 @@ def test_line_display_of_line() -> None:
 
     assert line_display == LineDisplay(
         line.number,
+        tuple(),
```
khoidt, line 22: I think it's a better practice to extend the line factory to generate old line numbers or at least empty tuples for that field.

PR #276, opened 2022-07-20
Add new genres

ejimsan:


ejimsan:
[Add new genres](https://trello.com/c/mbiNF7Jf/619-add-new-genres)

PR #287, opened 2022-09-02
Search lines by lemma

fsimonjetz:
Implements https://trello.com/c/gNGHD57K


ebl/corpus/infrastructure/queries.py
```diff
@@ -185,3 +185,31 @@ def join_text() -> List[dict]:
             },
         }
     ]
+
+
+def text_title_query() -> dict:
+    return {
```
khoidt, line 191: Use `join` instead of current method name

ebl/corpus/infrastructure/mongo_text_repository.py
```diff
@@ -215,6 +219,69 @@ def query_by_transliteration(
...
+                {"$count": "line_count"},
+            ]
+        )
+
+        return (
+            DictionaryLineSchema().load(
+                lemma_lines,
+                many=True,
+            ),
+            next(total_count, {}).get("line_count", 0),
```
fsimonjetz, line 282: Is this right/necessary to add default value?

ebl/corpus/web/chapters.py
```diff
@@ -77,3 +81,21 @@ def on_get(
...
+
+    @falcon.before(require_scope, "read:texts")
+    def on_get(self, req: falcon.Request, resp: falcon.Response) -> None:
+        try:
+            pagination_index = int(req.params["paginationIndex"])
+        except ValueError as error:
+            raise DataError("Pagination Index has to be a number") from error
+        resp.media = DictionaryLinePaginationSchema().dump(
+            self._corpus.search_lemma(
+                req.params["lemma"], pagination_index, Genre(req.params.get("genre"))
```
fsimonjetz, line 98: Can `Genre` take `None` parameter?

fsimonjetz: No, it doesn't. We need to check if `genre` is `None` before passing it to `Genre`.

ebl/corpus/infrastructure/mongo_text_repository.py
fsimonjetz: Apparently it _is_ necessary: If there are 0 documents, mongo returns nothing. In that case, we need to manually return 0.

PR #292, opened 2022-09-23
chapter info has textName of Text

yCobanoglu:


yCobanoglu:
[Trello Card](https://trello.com/c/RoMVkA7H/608-display-mode-for-corpus-search)

A new Field TextName has to be added for Chapter and used for ChapterInfo. TextName is a lookup of the TextName of Texts Collections from TextId of Chapter from Chapter Collections. Additionally ChapterInfo has to have TranslationLine not as string but as proper TranslationLine so it can be displayed as Markup on the Frontend just as in the Corpus. ChapterInfo is similar to Chapter but has some differences.

ebl/corpus/web/chapter_info_schema.py
```diff
@@ -2,13 +2,25 @@
...
 from ebl.corpus.web.chapter_schemas import ApiLineSchema
-from ebl.transliteration.application.line_schemas import TextLineSchema
+from ebl.transliteration.application.line_schemas import (
+    TextLineSchema,
+    TranslationLineSchema,
+)
+
+
+class ChapterInfoLineSchema(ApiLineSchema):
+    translation = fields.List(  # pyre-ignore[15]
```
khoidt, line 12: Replace with a function to match the type in the parent class.

khoidt (approved):

ebl/corpus/web/chapter_info_schema.py
yCobanoglu: function will not work because the type of the parent class is string.

PR #309, opened 2022-10-27
corpus cache

yCobanoglu:


ebl/corpus/web/alignments.py
```diff
@@ -24,6 +26,7 @@ def on_post(
         name: str,
     ) -> None:
         chapter_id = create_chapter_id(genre, category, index, stage, name)
+        self._cache.delete(str(chapter_id))
```
fsimonjetz, line 29: Is it obligatory to cast to `str`? Or can we use the `ChapterId` object as key?

yCobanoglu: chapter_id has a custom str method. Not sure what would be improved by using the object.

PR #314, opened 2022-11-04
Fragmentarium lemma search

fsimonjetz:


ebl/fragmentarium/web/fragments.py
```diff
@@ -21,7 +23,19 @@ def __init__(self, finder: FragmentFinder):
     @falcon.before(require_scope, "read:fragments")
     def on_get(self, req: Request, resp: Response, number: str):
         user: User = req.context.user
-        fragment, has_photo = self._finder.find(parse_museum_number(number))
+        lines: Union[Sequence[str], None] = req.params.get("lines")
+        fragment, has_photo = self._finder.find(
```
yCobanoglu, line 27: `req.params.get("lines", [])` so you don't have to check for `None` in the list comprehension.
For User input you have to validate that `int(i)` doesn't throw an error by `validating the params first. Catching the error and throwing a custom error ("Lines have to be numbers"). That has been done in some other place in the api.

ebl/fragmentarium/infrastructure/mongo_fragment_repository.py
```diff
@@ -180,10 +208,13 @@ def create_join(self, joins: Sequence[Sequence[Join]]) -> None:
...
         )
 
-    def query_by_museum_number(self, number: MuseumNumber):
+    def query_by_museum_number(
+        self, number: MuseumNumber, lines: Optional[Sequence[int]] = None
+    ):
         data = self._fragments.aggregate(
             [
                 {"$match": museum_number_is(number)},
+                *filter_fragment_lines(lines),
```
yCobanoglu, line 217: `lines and *filter_fragment_lines(lines)` works if `line is an empty list`

ebl/fragmentarium/infrastructure/mongo_fragment_repository.py
```diff
@@ -97,6 +105,26 @@ def _find_adjacent_museum_number_from_sequence(
     return current_prev, current_next
 
 
+def filter_fragment_lines(lines: Optional[Sequence[int]]) -> Sequence[dict]:
+    return (
+        [
+            {
```
yCobanoglu, line 111: it should be enough to return an item no need for list here i think. So `{...} `instead `[{...}]`.

ebl/fragmentarium/infrastructure/mongo_annotations_repository.py
```diff
@@ -29,7 +29,9 @@ def create_or_update(self, annotations: Annotations) -> None:
             True,
         )
 
-    def query_by_museum_number(self, number: MuseumNumber) -> Annotations:
+    def query_by_museum_number(
+        self, number: MuseumNumber, lines: Optional[Sequence[int]] = None
+    ) -> Annotations:
         try:
```
yCobanoglu, line 35: lines is not used why is it needed as parameter ?

ebl/tests/fragmentarium/test_fragment_finder.py
```diff
@@ -26,7 +26,11 @@ def test_find(
 ):
     fragment = FragmentFactory.build()
     number = fragment.number
-    (when(fragment_repository).query_by_museum_number(number).thenReturn(fragment))
+    (
```
yCobanoglu, line 29: `query_by_museum_number(number, lines) `needs a test

ebl/common/query/query_result.py
```diff
@@ -0,0 +1,20 @@
...
+from ebl.transliteration.domain.museum_number import MuseumNumber
+
+
+@attr.s(auto_attribs=True, frozen=True)
+class QueryItem:
+    id_: str
+    museum_number: MuseumNumber
+    matching_lines: Sequence[int]
+    total: int
+    lemma_sequences: Sequence[LemmaLine] = attr.ib(default=tuple())
```
yCobanoglu, line 14: Type should be tuple. Also `Tuple[LemmaLine] = tuple()`.

ebl/common/query/query_result.py
```diff
@@ -0,0 +1,20 @@
...
+
+
+@attr.s(auto_attribs=True, frozen=True)
+class QueryResult:
+    items: Sequence[QueryItem]
+    total_matching_lines: int
```
yCobanoglu, line 20: Also `Tuple`

ebl/common/query/query_result.py
yCobanoglu, line 14: what is the use of `total` maybe a more descriptive name like `total_`?

ebl/fragmentarium/web/fragments.py
yCobanoglu: Also need a test which tests "lines".

ebl/fragmentarium/infrastructure/phrase_matcher.py
```diff
@@ -0,0 +1,78 @@
...
+        line = list(
+            dropwhile(lambda unique_lemmas: self._phrase[0] not in unique_lemmas, line)
+        )
+
+        for ngram in ngrams(line, self._phrase_len):
+            if self._is_match(ngram):
+                print(">>> matches")
+                return True
+
+            print(">>> doesn't match")
```
yCobanoglu, line 37: Do you want to print this on the server logs ?

ebl/fragmentarium/web/fragments.py
```diff
@@ -21,7 +23,19 @@ def __init__(self, finder: FragmentFinder):
...
             check_fragment_scope(req.context.user, fragment.authorized_scopes)
         resp.media = create_response_dto(fragment, user, has_photo)
+
+
+class FragmentsQueryResource:
+    def __init__(self, repository: FragmentRepository):
+        self._repository = repository
+
+    def on_get(self, req: Request, resp: Response):
+        resp.media = QueryResultSchema().dump(self._repository.query_lemmas(req.params))
```
yCobanoglu, line 41: From what i understand req.params is supposed to be a dict with a single key and single value. Then it would be better to parse and validate it and pass to strings to `query_lemmas`.

ebl/fragmentarium/infrastructure/mongo_fragment_repository.py
```diff
@@ -439,3 +470,23 @@ def update_references(self, fragment):
...
+        cmd, _lemmas = next(iter(query.items()))
+
+        query_operator = QueryType[cmd.upper()]
+        lemmas = _lemmas.split("+")
+
+        data = self._fragments.aggregate(
+            create_search_aggregation(query_operator, lemmas)
+        )
+
+        if data := next(data, {}):
```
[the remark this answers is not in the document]
yCobanoglu:
```python
        if next(data, None):
            if query_operator == QueryType.PHRASE:
                filtered_data = filter_query_results(data, lemmas)
                return QueryResultSchema().load(filtered_data)
        else:
            return QueryResult([], 0)
```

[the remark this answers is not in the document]
yCobanoglu: If i understand this correclty you are assigning data to {} which leads it to jump to the else branch but in the else branch you are not using data so there is no point in using the walrus operator.

ebl/fragmentarium/infrastructure/mongo_fragment_repository.py
yCobanoglu, line 217: `lines and *filter_fragment_lines(lines)` works (even if line is an empty list)

ebl/fragmentarium/infrastructure/mongo_fragment_repository.py
```diff
@@ -97,6 +105,26 @@ def _find_adjacent_museum_number_from_sequence(
     return current_prev, current_next
 
 
+def filter_fragment_lines(lines: Optional[Sequence[int]]) -> Sequence[dict]:
+    return (
+        [
```
yCobanoglu, line 110: it should be enough to return an item no need for list here i think. So `{...} `instead `[{...}]`.

ebl/fragmentarium/infrastructure/mongo_annotations_repository.py
```diff
@@ -29,7 +29,9 @@ def create_or_update(self, annotations: Annotations) -> None:
             True,
         )
 
-    def query_by_museum_number(self, number: MuseumNumber) -> Annotations:
+    def query_by_museum_number(
+        self, number: MuseumNumber, lines: Optional[Sequence[int]] = None
+    ) -> Annotations:
```
yCobanoglu, line 34: lines is not used why is it needed as parameter ?

ebl/common/query/query_result.py
yCobanoglu, line 14: Tuple would be better. Also `Tuple[LemmaLine] = tuple()`. Same with matching_lines.

ebl/common/query/query_result.py
```diff
@@ -0,0 +1,20 @@
...
+
+from ebl.transliteration.domain.museum_number import MuseumNumber
+
+
+@attr.s(auto_attribs=True, frozen=True)
+class QueryItem:
+    id_: str
+    museum_number: MuseumNumber
+    matching_lines: Sequence[int]
+    total: int
```
yCobanoglu, line 13: what is the use of `total` maybe a more descriptive name like `total_`?

ebl/common/query/query_result.py
```diff
@@ -0,0 +1,20 @@
...
+    lemma_sequences: Sequence[LemmaLine] = attr.ib(default=tuple())
+
+
+@attr.s(auto_attribs=True, frozen=True)
+class QueryResult:
+    items: Sequence[QueryItem]
```
yCobanoglu, line 19: Type Tuple should work here too

ebl/fragmentarium/infrastructure/phrase_matcher.py
```diff
@@ -0,0 +1,78 @@
...
+        )
+
+    def matches(self, line: LemmaLine) -> bool:
+        line = list(
+            dropwhile(lambda unique_lemmas: self._phrase[0] not in unique_lemmas, line)
+        )
+
+        for ngram in ngrams(line, self._phrase_len):
+            if self._is_match(ngram):
+                print(">>> matches")
```
yCobanoglu, line 34: Probably you forgot to delete this debug information.

ebl/fragmentarium/infrastructure/phrase_matcher.py
```diff
@@ -0,0 +1,78 @@
...
+                return True
+
+            print(">>> doesn't match")
+        return False
```
yCobanoglu, line 38: Same here

ebl/fragmentarium/web/fragments.py
```diff
@@ -21,7 +23,19 @@ def __init__(self, finder: FragmentFinder):
     @falcon.before(require_scope, "read:fragments")
     def on_get(self, req: Request, resp: Response, number: str):
         user: User = req.context.user
-        fragment, has_photo = self._finder.find(parse_museum_number(number))
+        lines: Union[Sequence[str], None] = req.params.get("lines")
```
yCobanoglu, line 26: `req.params.get("lines", [])` so you don't have to check for `None` and just have one type. And later you would have to check if the list is empty.
For User input you have to validate that `int(i)` doesn't throw an error by validating the params first. Catching the error and throwing a custom error ("Lines have to be numbers"). Something similar has been done in the api for reference.

ebl/fragmentarium/web/fragments.py
yCobanoglu, line 27: Also need a test which tests "lines". A test in routes.

ebl/fragmentarium/web/fragments.py
yCobanoglu, line 41: From what i understand req.params is supposed to be a dict with a single key and single value. Then it would be better to parse and validate it and pass to strings to `query_lemmas`.

yCobanoglu (changes requested):

ebl/fragmentarium/infrastructure/mongo_fragment_repository.py
fsimonjetz: Not sure if I'm following - can you specify what you mean?

ebl/common/query/query_result.py
fsimonjetz: > Tuple would be better

I disagree: Tuples are for fixed-length, possibly heterogeneously typed sequences of elements. Here we have homogeneously typed sequences of varying length -> that's conceptually a list. For that we use `Sequence` type in the code base and cast into `tuple` for immutability.

ebl/common/query/query_result.py
fsimonjetz: It's basically `len(matching_lines)`, I'll rename it

ebl/fragmentarium/web/fragments.py
fsimonjetz: `None` -> "don't filter"/return all lines
`[]` -> "return no lines".

I.e., `None != []`

fsimonjetz: In principle you're right, but this is not for user input. But if we want to use this for user input at some point, we should add validation!

ebl/fragmentarium/web/fragments.py
fsimonjetz: You're right, I changed it. We will soon want to add additional filters, but handling them separate from the main query is probably a good idea.

ebl/fragmentarium/web/fragments.py
yCobanoglu: Api calls in principle should still be validated users could theoretically still call this api even if there is no frontend interface for it.

ebl/fragmentarium/infrastructure/mongo_fragment_repository.py
yCobanoglu: This makes only sense if `None` and `[]` represent the same but like you said empty list has a meaning it makes no sense. Otherwise `lines and *filter_fragment_lines` return either `lines` if `lines` is not empty or None or if its empty or None it return `filter_fragment_lines(lines)`.

fsimonjetz: Thanks for clarifying, I understand! (But it's the other way around, no? `lines` would be returned if it *is* empty or `None`, otherwise the filter gets called.) Either way, as you said it'd be a good solution if `None == []`. Do you think it would make sense to create a [sentinel value](https://death.andgravity.com/sentinels) for `None` to make it more obvious that it has a meaning to it?

yCobanoglu (approved):

PR #454, opened 2023-09-07
Add variant index positions

fsimonjetz:
Store the index of each variant in chapter.line.variants in a new `originalIndex` field for easier handling of partial ChapterDisplay in frontend corpus search results


fsimonjetz:
[Redesign Partial Chapter Display](https://trello.com/c/dz8LN5RV/714-redesign-partial-chapter-display)

PR #458, opened 2023-09-22
Add ngram endpoints

fsimonjetz:


ebl/corpus/infrastructure/mongo_text_repository.py
```diff
@@ -449,3 +432,90 @@ def query_corpus_by_manuscript(
             ]
         )
         return ManuscriptAttestationSchema().load(cursor, many=True)
+
+    def _update_ngrams(self, id_: ChapterId) -> None:
```
fsimonjetz, line 436: Needs to exclude colophons

PR #470, opened 2023-10-18
AfO Register

khoidt:


khoidt:
[Add AfO Register](https://trello.com/c/G3mhvOfq/717-add-afo-register)

PR #485, opened 2023-11-10
More informative error messages

fsimonjetz:


fsimonjetz:
[Show Cause of Duplicate Label Error](https://trello.com/c/4rirxFCP/50-show-cause-of-duplicate-label-error)

PR #515, opened 2024-01-30
Added isEnvelope property to class Join

Vas9ka:


khoidt (changes requested):
The coverage check looks good, but you still need tests for `is_envelope`. You need to ensure that the join tests (`ebl/tests/fragmentarium/test_joins.py`) include checks for this particular field. Also, it is important to extend `JoinFactory` (`ebl/tests/factories/fragment.py:96`) with `is_envelope` and update the tests that use it, if needed. As I noticed, there is another (duplicate) `JoinFactory` (`ebl/tests/factories/corpus.py:65`), which seems more up-to-date. I would suggest replacing the first factory's code with the second. The factory in `ebl/tests/factories/corpus.py:65` should be removed and imports updated.

Vas9ka:
[Add envelope to Joins collection](https://trello.com/c/yTCzUJVh/685-add-envelope-to-joins-collection)

khoidt (approved):

PR #534, opened 2024-03-06
Add reverse order to signs

ejimsan:


ejimsan:
[Add reverse order to sings](https://trello.com/c/qg7SOBL2/765-add-reverse-order-to-sings)

PR #543, opened 2024-04-11
Optimize Thumbnail Generation

fsimonjetz:


fsimonjetz:
Moved the thumbnail generation to the main image upload script managed by @ejimsan.

PR #548, opened 2024-04-26
Add Figulla Folios

ejimsan:


ejimsan:
[Import Figulla folios](https://trello.com/c/m59aadSC/772-import-figulla-folios)

PR #552, opened 2024-05-16
Update ATF importer

khoidt:


fsimonjetz (commented):
The edits are way too extensive for a single PR/review (we can keep discussing the exact line count, but anything above 400 LOC is [not reasonably reviewable](https://bssw.io/blog_posts/pull-request-size-matters)). The edits must be split into multiple PRs.

The Trello card suggests at least 8 conceptually distinct issues, each of which requires a dedicated PR.

1. Improve test coverage
2. Refactor massively to meet CodeClimate’s criteria
3. Import extend (e.g. `#tr.en.(o 2):`) in translation
4. Import CDLI and Oracc C-ATF
   1. Convert `'`, `~`, etc.
   2. Acceptable $- and \@-lines, e.g. `$ single ruling`,`@obverse`,`@column` ❓
5. Import Oracc
6. Instead of importing the whole edition and leave out lines that cannot be imported, prompt the user to correct lines that are incorrect.
7. Joins error

Other aspects that are not clear to me and likely require additional PRs include:

1. "If there is already an edition, a new edition cannot be imported." <- Is this the status quo and should be changed? Or is it a request?
2. Acceptable `@`-lines, e.g. `@obverse`, `@column` <- What about them?
3. Separate Lark grammar? <- What about it?

It is unclear to me which parts of the commit history map to which aspects of the Trello card. I tried to isolate changes regarding the Lark grammar and port them (#618), hoping this would get the bulk of the edits out of the way. However, I fail to see the need/benefit for a complete restructuring of the grammars. Specifically, I do
not support the proposed merging of the two grammars into a single one with disjunctions for legacy ATF items. This is guaranteed to cause issues down the line. The best practice in computational linguistics is to keep one grammar for one language, such that later changes in one grammar do not have unintended consequences for the other. After all, we would not incorporate, say, Dutch syntax
into a German grammar.

Although Lark allows for the definition of a shared base grammar and extensions for specific dialects, our grammar was not initially laid out like this. Restructuring it at this
point would require similarly extensive updates to the codebase as are in place in this PR. What do we gain from this?

In general, bug fixes should take precedence over restructuring, refactoring, and extensions ("Premature optimization is the root of all evil"). Ideally, each commit should clearly describe what the edit does, e.g., "add prime conversion", "fix transformer namespace". Avoid uninformative commit messages like "updates", "improve", etc. If larger parts of the code were edited, groups of lines that form logical units should be identified, staged and committed individually. As with PR sizes, smaller commits are generally better. Even a one-line edit can and often does form a
reasonable commit.

If merging an intermediate PR would introduce breaking changes, there are ways to deal with that, too (like [stacking PRs](https://blog.logrocket.com/using-stacked-pull-requests-in-github/)). It does not warrant oversized PRs.

Specific implementation issues include:

1. The code contains multiple formatting and type errors
2. There is a large number of comments and commented out code; it is not clear to me which of the  mentioned to-dos are already done and which ones are still open.
3. The tinkering with the namespaces is viable in principle; it even seems to be the conventional way in more recent version of lark (i.e., it is basically what the `merge_transformers` [function](https://github.com/lark-parser/lark/blob/87bb8efe0d425187159b39fd788640da33d1878e/lark/visitors.py#L190) does). However, it should be generalized into a higher-level function and applied to transformers where necessary instead of incorporating it into the `__init__` call. Better yet, we should upgrade Lark and just utilize the existing `merge_transformers` function.

.devcontainer/README.md
```diff
@@ -0,0 +1,66 @@
...
+   ```bash
+   cp .env.example .env
+   ```
+
+3. **Update Values**: Edit `.env` with your actual credentials for:
+   - Auth0 configuration (audience, issuer, PEM certificate)
+   - MongoDB connection URI and database name
+   - Sentry DSN
+   - AI API endpoint
+
```
khoidt:
```suggestion

### How Environment Variables are Loaded

- **Shell sessions**: The `.env` file is automatically loaded in new shell sessions via `~/.bashrc`
- **Task commands**: The `.env` file is loaded via the `dotenv: [".env"]` directive in `Taskfile.dist.yml`
- **Current session**: If you need to load `.env` in your current shell, run:
  ```bash
  set -a && source .env && set +a
  ```

```

ebl/tests/atf_importer/test_atf_importer.py
```diff
@@ -1,44 +1,605 @@
...
+def mock_input(monkeypatch):
+    def _set_input_responses(responses):
+        responses_iter = iter(responses)
+        monkeypatch.setattr(builtins, "input", lambda *args: next(responses_iter))
+        return responses_iter
+
+    return _set_input_responses
+
+
+def test_logger_writes_files(fragment_repository, tmp_path):
```
khoidt:
```suggestion
def test_logger_writes_files(fragment_repository, tmp_path, mock_input):
```

PR #564, opened 2024-07-14
Add Chicago numbers to Fragment

ejimsan:


ejimsan:
[Add Chicago ISAC to links](https://trello.com/c/3W0IqnBP/781-add-chicago-isac-to-links)

PR #601, opened 2025-02-17
Fix 500 errors on rapid saving

fsimonjetz:
Saving multiple times within a short period of time causes 500 errors due to rate limits of auth0. This is caused by the fact that saving editions performs 3 separate authentication calls to /userinfo, one for the notes, the introduction, and the transliteration, even if not changes were made. This PR adds a dedicated /edition endpoint to handle changes to these fields with a single API call, thus mitigating the rate limit issue.


fsimonjetz:
[Solve 500 errors on Saving Fragments](https://trello.com/c/kYhsFkDt/759-solve-500-errors-on-saving-fragments)

PR #636, opened 2025-09-29
Re train ocr model

JebediahC:
Merge code during retraining ocr model

## Summary by Sourcery

Improve write_fragment_numbers to better validate arguments, handle context initialization failures gracefully, and use dynamic repositories and output paths.

Enhancements:
- Validate that `output_annotations` and `output_imgs` are either both provided or neither using an XOR check
- Wrap full context creation in a try/except and fall back to direct Mongo client setup if it fails
- Retrieve `photo_repository` dynamically from the successful context path
- Construct the dated annotations output file path relative to `args.output_annotations`


ejimsan (commented):
@JebediahC can you please address the changes suggested by sourcery-ai (https://github.com/ElectronicBabylonianLiterature/ebl-api/pull/636).

Also, please adapt the file `/workspace/ebl-api/ebl/tests/fragmentarium/test_retrieve_annotations.py`. You should add new tests for:
- Argument parsing: A test where only one of --output_annotations or --output_imgs is passed should raise `argparse.ArgumentError`. A test where both are missing should set the defaults (`./annotations/...`).
- Context creation fallback. A test where `create_context()` raises an exception and the fallback Mongo setup is used (simulate this with `mockito`).
- Path handling for write_fragment_numbers: A test where you pass a custom `--output_annotations` directory and check that the file is written to its parent dir.

PR #711, opened 2026-04-30
Extend date metadata

khoidt:
Implement `isReconstructed` and `isEmended` fields in the `Year` class to enhance metadata handling. Introduce parsing logic to interpret raw year values, stripping wrapper symbols and setting corresponding flags. Ensure backward compatibility and maintain test coverage.

## Summary by Sourcery

Extend year metadata to support reconstructed and emended flags and introduce parsing of wrapper symbols into structured fields.

New Features:
- Add is_reconstructed and is_emended metadata fields to the Year domain model and schema.
- Support parsing of wrapper symbols and markers in year strings into structured metadata flags while normalizing the stored value.

Enhancements:
- Extend test factories and fragmentarium date tests to cover new year metadata, parsing behavior, and related date entities.
- Add task documentation files capturing requirements, work log, and TODOs for the date metadata changes.

Tests:
- Add and update unit tests for YearSchema parsing, new metadata fields, DateKing, and DateEponymSchema to ensure correct behavior and backward compatibility.


ebl/fragmentarium/domain/date.py
```diff
@@ -82,10 +84,62 @@ def filter_none(self, data: dict, **kwargs):
...
+        if _apply_wrapped_value_rule(data, value, start, end, key):
+            break
+
+    value = data["value"]
+    trailing_marker_rules = [("!", "is_emended"), ("?", "is_uncertain")]
+    for marker, key in trailing_marker_rules:
+        if _apply_trailing_marker_rule(data, value, marker, key):
+            break
+
+    return data
```
Fabdulla1: So from what I understand is that wrapper markers are evaluated first then trailing markers are second. A problem I see is that for something like <5>? the wrapper pass doesn't match that the string ends with ?. So I can see the trailing pass stripping ? and could set uncertainty, but wrapper parsing isn't re run. Then <5> is let with isUncertain true and without isReconstructed. 

I think it might be worth it to either normalize mixed markers fully or maybe reject mixed notation?

Fabdulla1 (changes requested):
Just need to address my comment for the mixed marker pattern. I also see that the `test_date.py` only covers single-marker patterns so it might be good to also add tests for mixed marker patterns.

PR #712, opened 2026-05-09
Display of pca analysis in sign display

asimniaz-ai:
Added fallback support for annotations without `pcaClustering` in centroid image queries.
When `centroids_only=true`, annotations with missing or null `pcaClustering` are now also returned instead of being excluded.

## Summary by Sourcery

Relax centroid annotation filtering and normalize script period filters when fetching cropped sign annotations.

Bug Fixes:
- Include annotations with missing or null PCA clustering data when requesting centroid-only annotations.
- Translate script period abbreviations to full period names before querying cropped sign annotations.


Fabdulla1 (changes requested):
So I noticed in `mongo_annotations_repository.py` that the centroids_only predicate now includes annotations where pcaClustering is null/missing (via the $or + $ifNull branch), and not just annotations with isCentroid=true. I think since pca_clustering is optional in the domain/schema this can return unclustered annotations for requests with centroids_only=true. This will weaken centroid-only semantics and could break consumers that expect centroid metadata. Maybe we can keep centroids_only strict to isCentroid=true or maybe rename the contract if this behavior is intentional?

Also, remove the unused helper in cropped_annotations.py:12-16 as this will prevent dead code residue and future drift.

ebl/signs/web/cropped_annotations.py
```diff
@@ -6,6 +6,8 @@
     CroppedAnnotationService,
 )
 
+ABBREV_TO_NAME = {period.value[1]: period.value[0] for period in Period}
+
 
 def period_name_from_abbreviation(abbreviation: str) -> str:
     for period in Period:
```
Fabdulla1: Unused helper left behind. Active execution path now uses dict lookup at `cropped_annotations.py` around line 51. Lets remove the unused helper or switch to one shared path.

Fabdulla1 (changes requested):
Overall, great, just one change as requested and another one:
so I think there is some missing end-to-end coverage for script abbreviation normalization. I think we need to have an explicit test for abbreviation input on the cluster images route. I see that the normalization logic was added in `cropped_annotations` around line 51 but the current route success test only passes a long period name in `test_sign_images_route.py` at line 232. Adding a test that sends an abbreviation and verifies the same expected result would protect against regressions if Period enum values or mapping shape change. After these changes, it'll be good to merge! Good job!

PR #717, opened 2026-05-26
Fix devcontainer to auto-create `.env` and inject Codespaces secrets

khoidt:
Automatically create the `.env` file from `.env.example` before container build and inject any matching Codespaces secrets found in the host environment. Update documentation to reflect this change and correct previous inaccuracies regarding automatic setup.

## Summary by Sourcery

Ensure devcontainer creation succeeds by auto-generating the .env file before Docker runs and documenting the new initialization and secret injection behavior.

New Features:
- Add a devcontainer initialization script that creates .env from .env.example and injects matching Codespaces secrets from the host environment before container startup.

Documentation:
- Update devcontainer documentation to describe the initializeCommand behavior, automatic .env creation, and Codespaces secret injection, and correct earlier inaccurate setup guidance.

Chores:
- Add task log and TODO documentation describing the devcontainer .env issue, root cause analysis, chosen solution, and validation steps.


Fabdulla1 (changes requested):
Overall fine, just some issues. For files `devcontainer.json` at line 15, `.devcontainer/init.sh`, the initializeCommand hard requires host bash. It would fail for hosts without bash in PATH. 

Additionally, for `.devcontainer/init.sh` the secret injection overwrites matching keys in .env unconditionally whenever host env has that key. Rebuilding can silently clobber intentional local config. It would be nice to preserve user-edited values to avoid destructive rebuild.

Fabdulla1 (approved):
Everything looks good! Just make sure to address the github comment about the unused import.

PR #722, opened 2026-06-15
Search optimization

Fabdulla1:
Optimize fragment search to support lightweight, paginated fragment summary responses with photo metadata, compact matching-line previews, and configurable count behavior.

**New Features**
-Add explicit-limit fragment query summary responses for /fragments/query.
-Return structured fragment summaries including description, date/dating data, genres, archaeology, references, projects, dossiers, match metadata, and compact matching-line previews.
-Expose hasPhoto and thumbnailPath metadata for fragments in search responses.
-Introduce dedicated fragment query summary/result domain models and schemas for rich search result responses.
-Add offset support for paginated fragment search results.
-Add count=exact|none|page support:
count=exact preserves exact matchCountTotal.
count=none skips exact total-count metadata.
count=page supports pagination via hasNextPage.

**Bug Fixes**
-Preserve exact total match counts for transliteration searches when an explicit result limit is applied.
-Preserve parser version fallback for matching-line previews.
-Avoid timeout-prone summary hydration behavior by limiting aggregation output before hydration.

### DIFF
--- a/ebl/common/query/parameter_parser.py
+++ b/ebl/common/query/parameter_parser.py
@@ -5,6 +5,8 @@ from ebl.transliteration.application.transliteration_query_factory import (
     TransliterationQueryFactory,
 )
 
+COUNT_MODES = ("exact", "none", "page")
+
 
 def parse_integer_field(field: str) -> Callable[[Dict], Dict]:
     def parse_integer(parameters: Dict) -> Dict:
@@ -21,6 +23,18 @@ def parse_integer_field(field: str) -> Callable[[Dict], Dict]:
             ) from error
 
     return parse_integer
+
+
+def parse_non_negative_integer_field(field: str) -> Callable[[Dict], Dict]:
+    parse_integer = parse_integer_field(field)
+
+    def parse_non_negative_integer(parameters: Dict) -> Dict:
+        parameters = parse_integer(parameters)
+        if parameters.get(field, 0) < 0:
+            raise DataError(f"{field} must be non-negative")
+        return parameters
+
+    return parse_non_negative_integer
 
 
 def parse_lines(lines: Sequence[str]) -> Sequence[int]:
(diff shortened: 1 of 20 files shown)


khoidt (changes requested):
Hey, this looks good, but please address the following before merging:

| # | Sev | Finding (location) | Why it matters | Fix |
| --- | --- | --- | --- | --- |
| F1 | Med | Dead code: `fragment_summary_projection()` (queries.py:24), `fragment_summary_projection_lightweight()` (queries.py:44) and module-level `load_fragment_query_result()` (mongo_fragment_repository_get.py:69) are never called | Duplicate the real projection (drift risk) and are inherently uncovered, so they break the 100% coverage gate | Delete all three; `query()` already uses the hydrating `_load_fragment_query_result` method and the `FRAGMENT_QUERY_SUMMARY_PROJECTION` dict |
| F2 | Med | `fragment_pattern_matcher.py` grew 230 -> 322 lines (mongo/queries/test files were already over on master; this one is pushed over by the PR) | Crosses the project's non-negotiable 250-line hard gate | Extract the projection/count helpers (`_result_projection`, `_items_projection`, `_match_count_total_projection`, `_has_next_page_projection`, `_count_pipeline`) into a companion module |
| F3 | Low | New branches uncovered: `compare_query_results` NotImplemented (query_result.py:26-27), `FragmentQuerySummary.__eq__` duck-typing (fragment_query_summary.py:94-95), `deserialize_site` non-dict (fragment_query_summary_schema.py:72), `_hydrate_fragment_query_item` NotFoundError (mongo_fragment_repository_get.py:330) | Hard gate requires 100% on touched lines | Add a test exercising each branch |
| F4 | Low | Count-metadata toggle is dead: `show_count_metadata` defaults False and `filter_count_metadata` strips the fields, but the pipeline hardcodes `_showCountMetadata: True`, so the filter only fires for `create_empty()` | `/fragments/query` returns `hasNextPage`/`isMatchCountTotalExact` for real queries but omits them for param-less/empty queries — two shapes for one endpoint | Either always emit the two fields, or actually honor the flag (and drop the misleading default) |
| F5 | Low | `_matching_line_preview` (mongo_fragment_repository_get.py:285) and `matching_line_preview_of` (fragment_query_summary.py:116) index `text.lines[index]` with matcher-computed indices, unguarded | IndexError -> HTTP 500 if an index ever exceeds the projected `text.lines` (schema/version drift); risk is low since index semantics are pre-existing | Guard the access (skip indices >= len) |
| F6 | Info | `_hydrate_fragment_query_item` (mongo_fragment_repository_get.py:299-300) reads `fragment["description"]`/`fragment["script"]` via brackets while every other field uses `.get()` | KeyError -> HTTP 500 if those fields are ever absent on a document | Use `.get()` with a sensible default for consistency |

PR #727, opened 2026-06-23
dev: added alias

Fabdulla1:
This PR implements server-owned bibliography identity and alias-based compatibility for partner bibliography workflows.

Partner create and duplicate-override now generate canonical eBL bibliography IDs server-side. Any partner-submitted ID is preserved as a `partner_id` alias instead of becoming the canonical ID. Bibliography entries now support `citationKey` and embedded aliases, and lookup can resolve records by canonical ID, citationKey, alias, normalized alias, or the new read-only resolver endpoint for raw slash/special-character aliases.

The PR also adds citationKey generation/persistence for partner-created records, non-unique lookup indexes for citationKey and aliases, collision handling for generated IDs/citationKeys/aliases, and tests covering canonical ID ownership, alias preservation, special-character aliases, citationKey collision suffixing, duplicate override behavior, and resolver access.

No duplicate merge/deprecation workflow, citationKey backfill, production migration, reference rewriting, `@bib{...}` rewriting, or frontend changes are included.

Remaining Work:
- `citationKey` backfill --> basically update existing data
- citationKey apply-mode migration --> After the dry-run report is reviewed, a separate migration/apply mode can persist citationKeys to existing records. This should include safety flags, batching, collision handling, and manual-review support
- Unique indexes after audit -->Unique constraints should only be added after a dry-run audit confirms there are no existing citationKey or alias collisions.
- Partner-scoped alias uniqueness --> The current alias model preserves partner-submitted IDs, but future work should decide whether aliases are globally unique or scoped by partner/source. This matters if multiple partners use the same local ID.
- Duplicate merge/deprecation workflow --> Need to address duplicates
- Frontend citationKey work?
- Update docs if need be

## Summary by Sourcery

Introduce server-owned canonical bibliography IDs, citationKey support, and alias-based lookup for partner bibliography entries, including a new resolve endpoint and repository indexing.

New Features:
- Allow partner bibliography entries to be created with server-generated canonical IDs while preserving partner-submitted IDs as aliases.
- Add citationKey generation, storage, and lookup for bibliography entries, including collision-safe suffixing.
- Expose a /api/v1/bibliography/resolve endpoint to retrieve entries by canonical ID, citationKey, or raw alias with scope-based access control.

Bug Fixes:
- Ensure unsafe partner IDs containing control characters are rejected without mutating the database.
- Prevent ambiguous or colliding aliases from being created by returning conflict responses and leaving existing data unchanged.

Enhancements:
- Extend bibliography lookup to resolve entries by canonical ID, citationKey, or alias with precedence rules favoring canonical IDs.
- Add partner-specific CSL and duplicate-override schemas, including aliases, to validate incoming payloads.
- Introduce partner identity utilities for normalizing partner IDs, generating canonical bibliography IDs, and producing collision-resistant citation keys.
- Update duplicate override workflows so partner overrides create canonical IDs, store partner IDs as aliases, and handle citationKey collisions.

Tests:
- Add repository index tests for citationKey and alias fields.
- Add tests covering alias normalization, special-character handling, ambiguity detection, and resolve endpoint behavior.
- Expand partner create and duplicate-override tests to cover canonical ID generation, alias persistence, citationKey collisions, and permission checks.

Chores:
- Initialize bibliography repository indexes at app bootstrap alongside existing collections.

### DIFF
--- a/docs/openapi/ebl-partner-bibliography.v1.yaml
+++ b/docs/openapi/ebl-partner-bibliography.v1.yaml
@@ -67,7 +67,7 @@ paths:
   /api/v1/bibliography/{idOrCitationKey}:
     get:
       summary: Get one bibliography entry
-      description: Read-only partner endpoint. Currently resolves by eBL bibliography `id`; citation-key lookup is reserved until stored citation keys exist. Requires `export:bibliography`.
+      description: Read-only partner endpoint. Resolves by eBL bibliography id, citation key, or trusted alias. The literal path segment `resolve` is reserved for `/api/v1/bibliography/resolve`; resolve that identifier with `/api/v1/bibliography/resolve?identifier=resolve`. Generated canonical ids are unaffected. Requires `export:bibliography`.
       security:
         - bearerAuth: []
       parameters:
@@ -120,6 +120,34 @@ paths:
             application/json:
               schema:
                 $ref: "#/components/schemas/DuplicateCandidateResponse"
+  /api/v1/bibliography/resolve:
+    get:
+      summary: Resolve a bibliography identifier
+      description: Read-only resolver for eBL bibliography ids, citation keys, or trusted aliases. This static route reserves the literal path segment `resolve`; a literal identifier equal to `resolve` must be resolved with `/api/v1/bibliography/resolve?identifier=resolve`. Generated canonical ids are unaffected. Requires `export:bibliography`.
+      security:
+        - bearerAuth: []
+      parameters:
+        - in: query
+          name: identifier
+          required: true
+          schema:
+            type: string
+          description: eBL bibliography id, citation key, or trusted alias to resolve.
+      responses:
+        "200":
+          description: Resolved bibliography entry.
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/PartnerBibliographyEntry"
+        "400":
+          description: Missing identifier query parameter.
+        "403":
+          description: Missing `export:bibliography` scope.
+        "404":
+          description: Bibliography entry not found.
+        "409":
+          description: Identifier is ambiguous or follows a redirect loop.
   /api/v1/bibliography/duplicate-candidates:
     post:
       summary: Check a proposed bibliography entry for likely duplicates
--- a/docs/openapi/ebl-partner-bibliography.v1.yaml
+++ b/docs/openapi/ebl-partner-bibliography.v1.yaml
@@ -33,7 +33,7 @@ paths:
           description: Missing `export:bibliography` scope.
     post:
       summary: Create a bibliography entry
-      description: Minimal partner create wrapper around existing bibliography create behavior. Requires `write:bibliography`. Duplicate detection runs before create; likely and possible duplicates return `409 Conflict`. Idempotency, external partner mapping persistence, and citationKey assignment are not implemented yet.
+      description: Create a partner bibliography entry. Requires `write:bibliography`. Duplicate detection runs before create; likely and possible duplicates return `409 Conflict`. The server assigns the canonical eBL bibliography id and citation key when enough citation data is present. A partner-submitted `id` is stored as a trusted alias, not honored as the canonical id. Idempotency and external partner mapping persistence are not implemented yet.
       security:
         - bearerAuth: []
       requestBody:
@@ -59,11 +59,13 @@ paths:
         "403":
           description: Missing `write:bibliography` scope.
         "409":
-          description: Duplicate candidate found; review before creating.
+          description: Duplicate candidate found; review before creating. Also returned when a submitted partner id or generated lookup value collides with an existing id, citation key, or alias.
           content:
             application/json:
               schema:
                 $ref: "#/components/schemas/DuplicateCandidateResponse"
+        "422":
+          description: Unprocessable partner payload, including server-owned fields, control characters in partner ids, or internal CSL schema validation failures.
   /api/v1/bibliography/{idOrCitationKey}:
     get:
       summary: Get one bibliography entry
@@ -89,7 +91,7 @@ paths:
           description: Bibliography entry not found.
     post:
       summary: Update a bibliography entry
-      description: Minimal partner update wrapper around existing bibliography update behavior. Requires `write:bibliography`. The path value is treated as an eBL bibliography `id`; citationKey lookup is not active until citation keys are stored. Duplicate detection runs before update while excluding the record itself. Idempotency and external partner mapping persistence are not implemented yet.
+      description: Update a partner bibliography entry. Requires `write:bibliography`. The path value is resolved by eBL bibliography id, citation key, or trusted alias using the same lookup rules as GET. Duplicate detection runs before update while excluding the record itself. Idempotency and external partner mapping persistence are not implemented yet.
       security:
         - bearerAuth: []
       parameters:
@@ -98,7 +100,7 @@ paths:
           required: true
           schema:
             type: string
-          description: eBL bibliography id for update.
+          description: eBL bibliography id, citation key, or trusted alias for update. The literal path segment `resolve` is reserved for `/api/v1/bibliography/resolve`.
       requestBody:
         required: true
         content:
@@ -115,11 +117,13 @@ paths:
         "404":
           description: Bibliography entry not found.
         "409":
-          description: Duplicate candidate found; review before updating.
+          description: Duplicate candidate found; review before updating. Also returned when the identifier is ambiguous.
           content:
             application/json:
               schema:
                 $ref: "#/components/schemas/DuplicateCandidateResponse"
+        "422":
+          description: Unprocessable partner payload, including server-owned fields or internal CSL schema validation failures.
   /api/v1/bibliography/resolve:
     get:
       summary: Resolve a bibliography identifier
@@ -217,12 +221,12 @@ components:
     BibliographyEntry:
       type: object
       required:
-        - id
         - type
-      additionalProperties: true
+      additionalProperties: false
       properties:
         id:
           type: string
+          description: Optional partner-provided id. On create, this value is stored as a trusted alias while the server assigns the canonical eBL bibliography id.
         type:
           type: string
         title:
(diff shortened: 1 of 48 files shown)


khoidt (changes requested):
Hi, reviewed and ran it locally — full review attached:

[TASK-727-review.md](https://github.com/user-attachments/files/30002789/TASK-727-review.md)

Direction is right, but I don't think it's mergeable yet. The main problem: only the submitted `id` is validated, while `aliases`, `deprecated` and `redirectTo` go straight from the partner payload into the DB. So a partner can poison an alias permanently (duplicate aliases → lookup raises forever), or POST an entry that redirects lookups to any record they choose — both reproduced. Also, `validate_references` now resolves aliases but the reference `$lookup` still joins on `_id`, so a reference can validate, save, and silently never hydrate.

Plus three hard gates failing (250-line limit, coverage, mypy). Details and repro steps in the review — happy to pair on the schema fix.

khoidt (changes requested):
Hi, looks good! The substantive security findings from the previous review were fixed and no longer reproduce. A few things still need a pass before I can approve. Three are project gates rather than bugs: Five test files exceed the 250-line hard limit, coverage on new code is below 100%, and two new `mypy` errors were introduced.

## Details:

- **250-line file limit — 5 test files over.** All under `ebl/tests/bibliography/`: `test_partner_bibliography_export_route.py` 168 -> 572, `test_bibliography.py` 210 -> 410, `test_bibliography_route.py` 219 -> 290, `test_partner_bibliography_override_success_route.py` 193 -> 278, `test_partner_bibliography_override_validation_route.py` 194 -> 253. Splitting the export-route file into resolve/alias, create and override modules would clear four of these at once. (`test_fragment_updater.py` at 411 was already over on master — not yours.)

- **Coverage on new code.** 98% locally, 93.6% on the diff; the gate is 100%. The gaps are the parts the PR exists to add: `partner_bibliography.py:147-153` and `:159` — the entire `DuplicateError` retry loop in `_create_prepared_entry` plus its exhaustion path, never exercised; `:166-167` — `_lookup_value_exists` treating `DuplicateError` as "exists", which is the guard stopping an already-ambiguous alias from being reused; `:191-192` — schema failure to `DataError`. Also `partner_identity.py:111` (`generate_partner_citation_key` returning `None` when creator/year/title are missing — plausible for real partner entries) and `:48` (empty normalized ID rejected). `bibliography.py:80` means `MAX_REDIRECT_DEPTH` is untested — the loop test trips `visited_ids` first.

- **Two new mypy errors.** `partner_bibliography.py:26` — `No overload variant of "frozenset" matches argument type "object"`, because `CSL_JSON_SCHEMA` is a bare dict so `["properties"]` types as `object`; `bibliography_entry.py` already solves this with `cast(dict[str, Any], ...)`. And `infrastructure/bibliography.py:101` — `query` is inferred `dict[str, str]` from its first assignment then reassigned to the `$or` form; annotate it `Dict[str, Any]`. CI only runs pyre and ruff, so these don't show up as a red check.

One correctness thing worth tracking rather than fixing here:

- **Alias uniqueness has a check-then-write race.** `_ensure_lookup_value_available` (`partner_bibliography.py:170`) checks, the insert happens at `:146`, and the alias indexes are created `unique=False` (`infrastructure/bibliography.py:77-82`). Two concurrent partner creates submitting the same ID both pass the check and both land. After that `query_by_alias` raises `DuplicateError` for that alias forever, so every lookup 409s and there's no way to repair it through the API. I know unique indexes are deliberately deferred until the collision audit — that's fine, could we just open a follow-up issue and think about a repair path?

And some small cleanups:

- `bibliography.py:212` — `validate_references` has no production callers now that `fragment_updater` uses `canonicalize_references`; it just discards the return value. Remove it or note why it stays.
- `bibliography_entries.py:26-28` — the hook builds `candidate_entries = [media]` then *replaces* it with `[media["bibliographyEntry"]]` instead of appending, so top-level fields are never inspected when `bibliographyEntry` is present. Not exploitable today (both partner schemas are `additionalProperties: false` and `@validate` runs right after), but it's the first layer and it doesn't do what it looks like it does.
- `infrastructure/bibliography.py:78` — `"aliases.value"` appears three times; qlty flags it (S1192). Worth a constant.
- Minor: `bootstrap.py:42` registers `/api/v1/bibliography/resolve` alongside `/api/v1/bibliography/{id_or_citation_key}`, and Falcon prefers the literal segment — an entry whose ID or `citationKey` is exactly `resolve` would be unreachable. Can't happen with `Q…` IDs, just noting it.

khoidt (approved):
Hi! Re-reviewed on the latest push, and the two commits from the 29th land really well. Everything substantive from the last round is genuinely fixed. Two things I'd still like before merging. One is the coverage gate — the retry loop that the whole reservation design exists for is the part with no test. The other is a small unrelated guard that snuck into the fragmentarium and can't actually execute. Everything else below is minor or just documentation catching up with the code.

### Details

**1. Coverage gate — 16 lines added by this PR are uncovered (blocking).** Full suite passes (3949 passed, 0 failures) and the bibliography package sits at 99%, but the gate is 100% on changed lines. The gaps are, unfortunately, the interesting ones:

- `partner_bibliography.py:152-164` — the entire `LookupValueInUseError` / `LookupValueReservationError` retry branch in `_create_prepared_entry`: canonical-id regeneration on a reservation conflict, citationKey regeneration on a citationKey conflict, and the `raise` for any other value. This is the concurrency path the reservation machinery was built for, and nothing exercises it.
- `partner_bibliography.py:197-206` — `_update_citation_key`, including the branch that drops `citationKey` entirely when regeneration returns `None`.
- `partner_bibliography.py:37, 40, 43, 46` — the `BibliographyCore` Protocol stub bodies. Easy one: `.coveragerc` already excludes `@abstractmethod`, so decorating these four clears them without touching any config.
- `bibliography.py:68` — deprecated entry with a missing or empty `redirectTo`.
- `lookup_identity.py:10` — the `continue` that skips a non-Mapping alias.
- `partner_identity.py:104` — `RuntimeError("Unable to generate bibliography id.")`.
- `bibliography_entries.py:24` — the early return in `reject_server_owned_partner_fields` when the media isn't a dict.
- `fragments.py:43` — see the next point.

(`bibliography_entries.py:87`, the `BibliographyList` cache-hit branch, is pre-existing and untouched — not yours.)

**2. `_parse_fragment_query` guard can't execute (`ebl/fragmentarium/web/fragments.py:41-45`).** The `if not callable(parser): raise DataError(...)` was added to silence CodeQL's "Non-callable called" alert, but the function has exactly one caller (line 123) and it passes module-level functions and closures — no code path can reach the raise. So it's dead, it's the uncovered `fragments.py:43` above, and it would convert a hypothetical `TypeError` into a 422 that can never be produced. It's also outside this PR's scope. I'd revert it and dismiss the alert as a false positive; if you'd rather keep it, it needs a test.

**3. The partner OpenAPI spec has drifted from the code in several places.** The backend schema is the source of truth, so the doc should follow it. The `GET` description was updated nicely, but:

- Line 36, `POST /api/v1/bibliography`: still says "citationKey assignment ... not implemented yet". It is implemented — the live response comes back with `"citationKey": "borger2003Zeichenliste"`.
- Same line: nothing says that a partner-submitted `id` is **not** honoured as the canonical id. This is the headline change of the PR — `POST` with `"id": "PARTNER/Bib 1"` returns `"id": "Q30000000"` and demotes the submitted value to an alias — and a partner reading only the spec would never find out.
- Line 92, `POST /api/v1/bibliography/{idOrCitationKey}`: still says "The path value is treated as an eBL bibliography `id`; citationKey lookup is not active until citation keys are stored." Update now goes through `Bibliography.find`, so it resolves by id, citationKey **and** alias.
- Lines 217-222, the `BibliographyEntry` request schema: declares `required: [id]` and `additionalProperties: true`, whereas `PARTNER_CSL_JSON_SCHEMA` requires only `type` and sets `additionalProperties: false`.
- Undocumented responses. `422` is now reachable on `POST` three separate ways (server-owned field, control-character id, internal schema failure) and the alias-collision `409` is distinct from the documented duplicate-candidate `409`. Only `400`, `403` and duplicate-candidate `409` are listed. All four verified live.

**4. `find_many` resolves canonical ids only, while `find` resolves aliases too.** `Bibliography.find_many` goes to `repository.query_by_ids`, which matches `_id`. So `GET /bibliography/{alias}` returns the entry but `GET /bibliography/list?ids={alias}` silently omits it — two read paths with two resolution rules. Related: `find_many` now de-duplicates after following redirects, so it can return fewer entries than ids requested. Both are defensible, but worth being deliberate about.

**5. An abandoned reservation being revived can surface as a spurious 404 (`lookup_reservations.py:156-161`).** `_handle_existing_reservation` re-reads the reservation and, if it's `abandoned`, calls `replace_one(...)` conditioned on it still being abandoned. The `deleteAt` TTL index (`expireAfterSeconds=0`, lines 35-37) can delete that document in between. `MongoCollection.replace_one` raises `NotFoundError` on `matched_count == 0`; `claim` doesn't catch it, it isn't a `DuplicateError`, and it maps to **404** — so a partner creating a brand-new entry could get "not found". `replace_one` already takes `upsert`, so passing `upsert=True` closes it.

**6. `datetime.utcnow()` in the new code.** `lookup_reservations.py:174`, `infrastructure/bibliography.py:44,65`, `bibliography_identity.py:31,42,65,72,74`, `web/bootstrap.py:21`. Deprecated since 3.12 and CI runs 3.12; `datetime.now(timezone.utc)` is the replacement. Minor related nit: `_pending_document` stamps `createdAt` from its own `utcnow()` even though the caller already threaded a `now` through, so one document carries two clocks.

**7. `_handle_existing_reservation` takes 6 parameters (`lookup_reservations.py:135`).** qlty's `function-parameters` rule fires at 6 in this repo — it's what it flagged in the fragmentarium tests. New production code, so probably worth folding `operation`/`value`/`now` together.

**8. Three pre-existing `mypy` errors in files this PR touches.** Not introduced here, but the gate covers touched files: `test_bibliography_query_builders.py:36` (`Cannot assign to a method`), `test_fragment_updater.py:18` (`lark_parser has no attribute parse_atf_lark` — PR #743 is fixing this repo-wide), `test_app_bootstrap.py:66` (`ebl_ai_client` is read-only).

Nice work on the reservation design — it's a real improvement over what was there.

PR #733, opened 2026-07-01
fix: pagination issues

Fabdulla1:
Add real pagination support to `/fragments/query`. Avoid returning all matching fragments in one large response for large genre searches.

## Summary by Sourcery

Implement proper pagination and total count reporting for fragment query endpoints and ensure related infrastructure scales to large genre searches.

Bug Fixes:
- Fix fragments query endpoint returning all matching fragments by adding offset/limit-based pagination and ignoring deprecated paginationIndex.
- Reject invalid negative or non-integer pagination offsets at the API layer.

Enhancements:
- Include totalCount in query results alongside matchCountTotal to expose overall result size independent of pagination.
- Restructure the fragment search aggregation pipeline to use a facet that simultaneously paginates items and computes totalCount.
- Allow MongoDB aggregations for fragment queries to use disk to handle large datasets.
- Add a genres.category index to improve performance of genre-based fragment queries.
- Default QueryResult.total_count to the number of items when not explicitly provided.

Tests:
- Extend fragment search route tests to cover totalCount, pagination behavior, invalid offsets, and ignoring paginationIndex.
- Add repository-level tests for genre querying with pagination and aggregation options.
- Introduce tests for the fragment query aggregation pipeline to verify pagination stages and metadata facets.
- Add tests for KMZ coordinate import utilities and safety filters around applying coordinate updates.
- Add tests ensuring bibliography routes correctly handle IDs containing non-ASCII characters.


khoidt (approved):
Hey! This is in good shape. Tests pass (80/80), the changed lines are fully covered, flake8 is clean, and there's nothing broken. Approving.

One thing worth calling out: the PR description promises a big pagination overhaul, but the actual diff is much smaller now — the pipeline work (`fragment_pattern_matcher.py`, `parameter_parser.py`, `query_result.py`, KMZ import) came off this branch and landed on `master` separately. That means most of the bot comments (Sourcery, qlty, Codex) point at files that aren't in the diff anymore, so they're safe to ignore. The one Codex comment that does touch live code — "offset-only should return empty" — is actually the opposite of what the tests intend, so no change needed there either.

What's genuinely here: stripping the deprecated `paginationIndex` (the real fix — before, a `paginationIndex`-only request silently did a full unfiltered scan), a `genres.category` index, `allowDiskUse=True`, plus CI/Docker install retries.

**Pending items — none blocking:**

1. **File-size gate.** `test_fragment_repository.py` (1520 lines) and
   `test_fragments_search_route.py` (1084) are both way over the project's 250-line limit, and this PR makes them bigger. Pre-existing, but worth splitting.
2. **Scope creep.** The CI/Dockerfile retries and the non-ASCII bibliography ID tests have nothing to do with pagination. Fine to keep, just maybe retitle or split.
3. **Heads-up, not a bug.** A request with only `offset`, `limit`, or `count` returns *all*
   authorized fragments, paginated. That's pre-existing and now locked in by tests — just be aware it can be a big response.

PR #735, opened 2026-07-03
Add GET /realia/all endpoint for listing Realia IDs for the sitemap

khoidt:
Add `GET /realia/all`, returning a sorted JSON array of Realia `_id`s for frontend sitemap generation. `ebl-frontend` already calls it from `RealiaRepository.listAllRealia` (used by `src/router/sitemap.tsx`); backend master has no listing route, so that call currently 404s.

## What the endpoint returns

Every ID it returns can be fetched back from `/realia/{id}`. Excluded:

- **Redirect stubs** — the same rule as the frontend's `getRedirectTarget`: exactly one cross-reference and no own content, where own content means a non-empty `afoRegister`, `references` or `afoCrossReferences`, more than one `reallexikon` entry, or a `reallexikon` entry with a resolvable reference. `relatedTerms`, `type` and `wikidataId` are metadata, not content. Entries with two or more cross-references are listed (the frontend renders them rather than redirecting).
- **Reserved identifiers** — `all` collides with the route itself. `ebl/realia/domain/reserved_identifiers.py` is the single source of truth for the route path and the exclusion. The API has no Realia write path, so an entry named `all` cannot be rejected at creation; it would be reachable only via `/realia/by-id/{realiaId}`. This is an accepted constraint.
- **Entries the detail route cannot load** — non-string `_id`s are filtered in Mongo; every remaining candidate is loaded through `RealiaEntrySchema` (the same load `/realia/{id}` performs) and skipped if it fails, e.g. an explicit `null` in a list field or in `realiaId`, or elements of the wrong shape. One malformed document can no longer fail the whole endpoint.

Results are sorted case- and accent-insensitively, with the raw ID as a tie-breaker.

## Caching and performance

The stub filter is a `$expr` guarded with `$isArray`/`$cond`, so it cannot use an index; candidates are then schema-validated and sorted in Python. The result is memoized server-side through the app cache (`cache.memoize`, 600 s) and the response carries `Cache-Control: public, max-age=600`, so the scan runs at most once per cache period when a cache backend is configured (`CACHE_CONFIG`; the default is the null backend). `cache.cached` is deliberately not used: on a cache hit it skips the responder hooks and drops the `Cache-Control` header. The response is unbounded by design, since a sitemap needs the complete list.

## Also in this PR

- The `/realia/{…}` URI template variable is renamed `realia_id` → `entry_id`, to distinguish the raw `_id` from `realiaId` (behaviour unchanged).
- `RealiaSearchResource` coerces a missing `query` to `""` (`or ""`) so the type checkers see a `str` (behaviour unchanged).
- The README's Caching section documents the `cache.cached` / `Cache-Control` interaction and the `cache.memoize` alternative.

## Summary by Sourcery

Provide the frontend sitemap with a reliable, cached listing of valid Realia IDs.

New Features:
- Add `GET /realia/all` to return a complete, sorted list of sitemap-eligible Realia IDs.

Bug Fixes:
- Exclude redirect stubs, reserved identifiers, malformed IDs, and schema-invalid entries so every returned ID remains retrievable.

Enhancements:
- Add robust redirect filtering and case- and accent-insensitive ID ordering.
- Memoize listing results for 600 seconds while preserving cache-control headers on every response.
- Rename the Realia route parameter for clarity and normalize missing search queries.

Documentation:
- Document the interaction between cached responders, cache-control headers, and memoized data in the README.

Tests:
- Add coverage for listing behavior, sorting, filtering, malformed documents, retrievability, and server-side caching.


Fabdulla1 (changes requested):
Everything looks good except in `ebl/realia/web/bootstrap.py` lines 20-22 the static `/realia/all` route seems to shadow a possible Realia entry whose `_id` is `"all"`. I don't see any evidence that such an entry exists but it still isn't reserved. I also think not every Realia entry is guaranteed to have a non-empty `reaiaID` for access through `/realia/by-id{realiaID}`. Maybe we just reserve `"all"` as an `_id` or document the route collision as an accepted constraint?

ebl/realia/web/bootstrap.py
```diff
@@ -14,7 +15,9 @@ def create_realia_routes(api: falcon.App, context: Context) -> None:
     realia_resource = RealiaResource(context.realia_repository)
     realia_by_id_resource = RealiaByIdResource(context.realia_repository)
     realia_search_resource = RealiaSearchResource(context.realia_repository)
+    realia_list_resource = RealiaListResource(context.realia_repository)
     realia_lemma_sink = RealiaLemmaSink(context.realia_repository)
+    api.add_route("/realia/ids", realia_list_resource)
     api.add_route("/realia/by-id/{realia_id}", realia_by_id_resource)
     api.add_route("/realia/{realia_id}", realia_resource)
     api.add_route("/realia", realia_search_resource)
```
Fabdulla1, line 23: changing the endpoint from `/realia/all` to `/realia/ids` makes a Realia entry whose raw `_id` is `"all"` reachable againt but it also creates the same collision for an entry whose `_id` is `"ids"`. Since `"ids"` is not reversed and `realiaID` may be blank or missing, `/realia/by-id/{realiaID}` is not a guaranteed fallback. It might be good to enforce `"ids"` as a reserved identifier across every creation and import path then maybe provide a guaranteed non-colliding raw-ID route? Or maybe we can use a route structure that can't overlap with Realia identifiers?

ebl/realia/infrastructure/realia_stub_filter.py
```diff
@@ -0,0 +1,73 @@
...
+def _array_size(field: str) -> dict:
+    return {"$size": {IF_NULL: [f"${field}", []]}}
+
+
+def _resolvable_reallexikon_count() -> dict:
+    return {
+        "$size": {
+            "$filter": {
+                "input": {IF_NULL: ["$reallexikon", []]},
+                "cond": _is_resolvable_reference("$$this.reference"),
```
Fabdulla1, line 52: The stub predicate uses `$size` and `$filter` after `$ifNull`. This handles missing and null values but it doesnt seem to protect against legacy documents where an expected array field contains a scalar or object. I'm thinking that in that case on malformed document could cause the entire `/realia/ids` query and also the sitemap generation to fail. I think it would be best to gaurd these expressions wth `$isArray`/`$cond` or maybe coerce non-array values to an empty array and then add coverage for malformed legacy field shapes?

Fabdulla1 (changes requested):
Overall, looks great, but I think there are some changes that need to be done. Hopefully, my comments are helpful!

Fabdulla1 (changes requested):
Hey! So, `realia_document_shape.py` seems to treat both `"missing"` and `"null"` as acceptable absent values but then `RealiaEntrySchema` doesn't accept explicit `null` for these array fields.

It might be good to treat only `"missing"` as absent here and exclude the explicit `null` values so it doesn't break the endpoint's central invariant that it shouldn't advertise entries that the detail route can't load. 

Aside from that, I also don't think the current `Cahce-Control: public, max-age=600` implementation supports the PR description's claim that caching absorbs repeated calls. 

`cache_control(...)` seems to only add HTTP response header and the application cache defaults to the null backend where this responder isn't wrapped with `cache.cached(...)`. I also see a deployment level CDN/reverse-proxy cache here so a direct request still performs the full Realia collection `$expr` scan, materializes the IDs, and sorts them in python every time. Since it's a public endpoint, it might be good to use the existing server-side caching pattern already used such as `cache_control(...) + cache.cached(...)` or update the PR description. 


Last but not least, would you be able to update the PR before merging. It seems the redirect/filter semantics have now changed substantially during the review cycle and the current body now describes behavior the final code doesn't really implement anymore. Thanks! 

Looking forward to this :D

PR #738, opened 2026-07-08
Feature media architecture

Fabdulla1:
## Summary by Sourcery

Introduce a persistence-agnostic media domain model and application-layer contracts, along with DTO schemas and tests, to support future media storage, import, and fragment-scoped APIs without changing existing routes or persistence.

New Features:
- Define a UUID-based media domain model with types, associations, representations, checksums, references, and import-source metadata independent of MongoDB/GridFS.
- Add application-layer interfaces for media repositories, representation storage, services, importers, and backfill operations to decouple future infrastructure from domain logic.
- Provide Marshmallow DTOs and serialization schemas for fragment media responses and summaries that preserve current hasPhoto/thumbnailPath semantics.

Architecture and operational contracts:
- Keep architecture and operational rules next to their contracts in module and class docstrings, including MIME and SVG safety, future URL construction and routing limitations, storage-version semantics, authorization boundaries, import/backfill requests, and media selection.
- Add no standalone Markdown documentation files.

Review follow-up:
- Canonicalize backfill reports without losing repeated-category findings, and make equality and hashing independent of construction order.
- Reject malformed or mixed request collections, negative report counters, invalid cursors, oversized batches, and invalid media-domain runtime types with consistent `ValueError` failures.
- Validate representation checksum types so malformed values cannot bypass the original-representation checksum invariant.
- Document Falcon 3.1.3's `%2F` route-matching limitation without changing URL output.

Tests:
- Add unit tests for the media domain model to enforce validation rules, association behavior, and deterministic ordering.
- Add tests for media DTO schemas to verify fragment-context serialization, field omission rules, and compatibility with legacy flags.
- Add application-contract tests using in-memory implementations to pin expected behavior of the media repository and service interfaces.
- Remove two exact duplicate `BackfillRequest` tests from the representation-store suite; canonical and expanded coverage remains in `test_media_request_validation.py`.


khoidt (changes requested):
Hi! The domain model and the scope discipline look good, and I can see all the earlier bot feedback (Sourcery, Codex, qlty) has already been dealt with.

Requesting changes for now. There are three hard-gate failures (the 250-line file limit, mypy, and coverage on new lines) which are all mechanical, plus one thing I'd like to settle before this lands: `hasPhoto` and `thumbnailPath` can disagree in `FragmentMediaSummaryDto`, which matters for the backfill.

Full review:

[TASK-738-review.md](https://github.com/user-attachments/files/30005885/TASK-738-review.md)

khoidt (changes requested):
Nice work — this is a big improvement over the last round. Every one of the six findings from the previous review is genuinely fixed, not just papered over: the four oversized files are split, the markdown lint error is gone, both mypy errors are cleared, coverage on the new code is a clean 100%, and the `hasPhoto` / `thumbnailPath` drift is properly resolved by making `thumbnailPath` unconditional again — with four tests pinning it so it can't quietly regress. All 19 bot threads (Sourcery, Codex, qlty, CodeQL) are resolved and I re-checked each one against the actual head rather than trusting the resolved flag. Every gate passes locally and CI is fully green.

The domain model is a pleasure to read. Invariants live in the constructors, the MIME policy around SVG is tight and well tested, and the abstract contracts are documented well enough that whoever writes the Mongo adapter will know exactly what's expected of them — the notes on logical versioning and the IDOR boundary are especially good.

Two things before merge. First, the PR adds two new `.md` files under `docs/`, which we don't want in the repo. Second, I found five small contract gaps in the not-yet-implemented importer/backfill request objects — none of them can bite today, but they're much cheaper to settle now, while it's all still on paper, than after the implementation is written against them.

Verdict: request changes for the two new docs; the rest is advisory and I'd be happy to see it land right after.

## Details

### Finding 1 — Two new `.md` files added to the repository — BLOCKING

The PR adds two documents:

| File | Lines |
| --- | --- |
| `docs/media-architecture.md` | 428 |
| `docs/media-importer-backfill-security.md` | 318 |

No new `.md` files should be present in the repository. Both files lint cleanly and their *content* is genuinely good — the domain model description, the future Mongo document shape, the authorization model, and the SVG security rules are all accurate against the code, and I found no stale references (nothing mentions the deleted `converters.py`). The objection is to the files existing in the tree, not to what they say.

Options, in order of preference:

1. Move the content out of the repository (wiki, issue, or the PR description) and delete both files.
2. Fold the parts that are genuinely load-bearing — the SVG-only-as-COPY-original rule, the logical-version handle contract, the IDOR boundary on `find_in_fragment` — into the module docstrings where they already almost live, and delete the files. Much of `media_store.py` and `media_repository.py` already carries this information inline, so the overlap is real.

Note that `docs/media-architecture.md:229-231` is the *only* place the legacy-vs-new URL encoding policy is written down (see Finding 7); wherever the docs end up, that paragraph should survive.

**Severity**: High · **Blocking**: Yes

---

### Finding 2 — `BackfillReport` is unhashable despite being a frozen attrs class — LOW

`ebl/media/application/media_requests.py:98-116`

`BackfillReport` is declared `@attr.s(auto_attribs=True, frozen=True)`, so attrs generates a `__hash__`. But `reports` is converted to a `MappingProxyType`, which is not hashable, so the generated `__hash__` raises at runtime. The class advertises a capability it does not have. Every other frozen value object in this PR — `Media`, `MediaId`, `StoredMedia`, `ImportReport`, `ImportRequest` — is genuinely hashable, so this is the odd one out.

The `MappingProxyType` is the right call for immutability; the fix is just to be explicit about the consequence — either `@attr.s(..., frozen=True, hash=False)`, or store `tuple[tuple[BackfillCategory, tuple[str, ...]], ...]` and expose the mapping through a property.

#### Reproduction

```python
from ebl.media.application import BackfillReport
hash(BackfillReport(reports={}))
# TypeError: unhashable type: 'mappingproxy'
```

**Severity**: Low · **Blocking**: No

---

### Finding 3 — `BackfillRequest` skips the validation discipline used everywhere else — LOW

`ebl/media/application/media_requests.py:91-95`

```python
@attr.s(auto_attribs=True, frozen=True)
class BackfillRequest:
    dry_run: bool = True
    batch_size: Optional[int] = None
    resume_after: Optional[str] = None
```

Three fields, no validators — in a module where `ImportRequest` right above it validates `mode` with `instance_of` and `source_name` with `not_blank`, and where the domain layer applies `positive_int`, `not_blank` and `strict_bool` consistently.

`MediaBackfill.backfill` documents these as preconditions — "Processes at most one bounded batch" and "the opaque cursor to pass back" — but nothing enforces them:

```python
BackfillRequest(batch_size=0)     # -> accepted; a batch that processes nothing
BackfillRequest(batch_size=-5)    # -> accepted; meaningless
BackfillRequest(resume_after="")  # -> accepted; an empty "opaque cursor"
BackfillRequest(dry_run="yes")    # -> accepted; truthy string, and dry_run is the safety switch
```

The last one is the one I'd actually fix: `dry_run` is the guard that stops a backfill mutating production data, and `MediaAssociation.is_primary` already uses `strict_bool` for a far less consequential flag. Suggested: `positive_int` on `batch_size` (optional), `not_blank` on `resume_after` (optional), `strict_bool` on `dry_run`. The same `strict_bool` argument applies to `ImportRequest.dry_run` at line 78.

**Severity**: Low · **Blocking**: No

---

### Finding 4 — `ImportRequest.fragment_ids` accepts duplicates — LOW

`ebl/media/application/media_requests.py:75-77`

```python
ImportRequest(ImportMode.REPLACE, "s", (MuseumNumber.of("K.1"), MuseumNumber.of("K.1"))).fragment_ids
# -> (MuseumNumber('K','1',''), MuseumNumber('K','1',''))
```

`_museum_numbers_of` only tuple-ifies. This sits awkwardly beside two neighbouring decisions that both went the other way: `Media._validate_associations` explicitly rejects duplicate fragment associations, and `MediaReader.find_by_fragments` explicitly documents "Duplicate inputs yield one key". With `ImportMode.REPLACE` a duplicated id means the same source is replaced twice in one run, and `ImportReport.replaced` double-counts it. Dedupe in the converter (`tuple(dict.fromkeys(...))`) or reject duplicates, and say which in the docstring.

**Severity**: Low · **Blocking**: No

---

### Finding 5 — Selection helpers raise on unassociated media, and that propagation is unpinned — LOW/MEDIUM

`ebl/media/application/media_selection.py:26-28, 47-49` and `ebl/media/application/media_summary_dtos.py:61-71`

`has_photo` and `fragment_media_in_order` both route through `Media.association_for`, which raises `ValueError` when the media is not associated with the fragment. So the summary DTO inherits a hard precondition — every item must be associated with `fragment_id` — that is documented nowhere on `FragmentMediaSummaryDto.of`, and is not pinned by a test at this layer:

```python
photo = photo_media(associations=(association(sort_order=0, is_primary=True),))  # K.1 only
has_photo(MuseumNumber.of("K.2"), (photo,))
# ValueError: Media 550e8400-... is not associated with fragment K.2.
```

Failing loudly is a defensible choice, and the repository contract does guarantee scoped reads. But the summary DTO is the piece that will be called per-row inside a fragment *query* response, where the media arrives via `find_by_fragments` as a mapping — one mis-keyed batch turns a whole search page into a 500 rather than one degraded summary. Right now the only test covering the raise is the domain-level one at `test_media_domain.py:228`; coverage is 100% because the happy path executes the same lines.

Two things worth doing, both cheap: state the precondition in `FragmentMediaSummaryDto.of`'s docstring, and add a test at the selection/DTO layer pinning what happens when it is violated — whichever behaviour you decide is correct.

Related nit, same file: `_fragment_media_type` (line 47) calls `media.association_for(fragment_id)` purely for the side effect of raising and discards the result. It reads as a leftover. `assert media.is_associated_with(...)` or a comment would make the intent survive the next reader.

**Severity**: Low to Medium · **Blocking**: No

---

### Finding 6 — New dependency edge: `fragmentarium` → `media` — INFORMATIONAL

`ebl/fragmentarium/application/fragment_query_summary_schema.py:20-21`

The single behavioural change in the PR replaces a hardcoded f-string with `legacy_fragment_thumbnail_url`, which introduces a new import from `ebl.media.application` into `ebl.fragmentarium.application`.

I checked and this is benign: `media_urls.py` is a leaf that imports only from `ebl.media.domain` and `MuseumNumber`, so there is no cycle, and CodeQL agrees (the earlier cyclic-import alerts referred to the since-deleted `ebl/media/domain/converters.py`). The output is byte-identical — `test_media_legacy_compatibility.py` has four tests pinning it, including the `MuseumNumber("A/B", "1")` edge case. Flagging only because it is the first edge from `fragmentarium` into `media`, and it is worth being deliberate about the direction before the routes land.

**Severity**: Informational · **Blocking**: No

---

### Finding 7 — Legacy `thumbnailPath` is not percent-encoded — INFORMATIONAL, pre-existing

`ebl/media/application/media_urls.py:33-36`

`legacy_fragment_thumbnail_url` interpolates the museum number raw, while all three `fragment_media_*_url` helpers encode each segment through `_path_segment`. For a museum number containing a slash the two disagree:

```python
legacy_fragment_thumbnail_url(MuseumNumber("A/B", "1"), ThumbnailSize.SMALL)
# '/fragments/A/B.1/thumbnail/small'   <- extra path segment, will not match a {museum_number} template
```

This is deliberate, documented at `docs/media-architecture.md:229-231`, byte-identical to master, and pinned by `test_production_summary_keeps_the_raw_legacy_thumbnail_path`. **Not a regression and not this PR's bug.** Recording it because the asymmetry is now load-bearing: when the real routes land, the legacy path will not round-trip through a Falcon `{museum_number}` template for these museum numbers, and whoever migrates that public contract will need this note. Keep the paragraph wherever the docs end up (Finding 1).

**Severity**: Informational · **Blocking**: No

---

### Finding 8 — Sourcery's route-drift concern is deferred, not resolved — INFORMATIONAL, carry-over

`media_urls.py` builds route strings that no registered Falcon route serves yet. Centralising them (which is what Sourcery asked for, and it was done) removes the duplication but does not remove the drift risk — it just moves it to one place. When the routes are registered, add a test asserting the registered route templates and these builders agree. Carried over from the previous review; repeating it so it does not get lost when these task files are deleted.

**Severity**: Informational · **Blocking**: No

khoidt (changes requested):
Good progress. Everything from my last round is handled: both `.md` files are gone with the important rules moved into module docstrings, the request objects validate their inputs, duplicate fragment ids are collapsed, and the selection precondition is stated and pinned by tests. All gates are green, every `ebl/media` module is at 100%, and I checked the one bit of this PR that is live today — `thumbnailPath` in the fragment query response — by running this branch and master side by side against the same data: the responses are byte-identical, including museum numbers with a slash and a space.

One regression came in with the fix for my `BackfillReport` finding, and it needs fixing: switching from a mapping to `(category, entries)` pairs means that if the same category appears twice, one set of findings silently vanishes — and two reports with the same findings in a different order no longer compare equal, which is the opposite of what the docstring promises. The other must-fix is `ImportRequest.fragment_ids`: it's typed as museum numbers but happily takes strings, mixed in with museum numbers, and splits a bare string into characters.

The rest are small validation gaps in contracts that nothing calls yet — cheap to settle now, before the importer is written against them — plus a stale PR description.

No dev container or config changes at head. For the record, the branch history did touch `.github/workflows/main.yml` (changed in `922b6249`, reverted exactly in `f5039397`) and picked up `Dockerfile` / `pyproject.toml` / `poetry.lock` only through merges from master; the net diff against master contains none of them. No `.md` files at head.

### Details

#### M1 — Medium — `BackfillReport` drops findings when a category repeats, and its equality depends on input order

`ebl/media/application/media_requests.py:63-70` (converter) and `:150-151` (`reports`), introduced in `8eab5e6f` as the fix for my round-2 Finding 2.

```python
def _report_entries_of(value):
    return tuple(sorted(((category, tuple(entries)) for category, entries in value or ()),
                        key=lambda entry: entry[0].value))

@property
def reports(self) -> Mapping[BackfillCategory, Sequence[str]]:
    return MappingProxyType(dict(self.report_entries))
```

The old mapping made a repeated category impossible; pairs allow it, and `dict(...)` keeps only the last one. `sorted` is stable, so pairs with the same category keep their input order, which is then what decides equality. Verified:

```text
a = BackfillReport(report_entries=((UNKNOWN_FRAGMENT, ("K.1",)), (UNKNOWN_FRAGMENT, ("K.2",))))
b = BackfillReport(report_entries=((UNKNOWN_FRAGMENT, ("K.2",)), (UNKNOWN_FRAGMENT, ("K.1",))))
dict(a.reports)  -> {UNKNOWN_FRAGMENT: ('K.2',)}     # K.1 lost
a == b           -> False                            # docstring: "equal and equally hashed whatever order they were built in"
```

An importer that appends one pair per finding — the natural way to build it — would lose every finding but the last in each category. A non-enum category also fails with `AttributeError` (`.value`) rather than a validation error. Fix: in the converter, merge entries per category (and sort the entries within a category), and require each key to be a `BackfillCategory`; add tests for the repeated-category and reversed-order cases.

#### M2 — Medium — `ImportRequest.fragment_ids` accepts strings, mixed with museum numbers, and splits a bare string

`ebl/media/application/media_requests.py:53-56, 88-90`.

The field is declared `Sequence[MuseumNumber]`, but nothing enforces the element type:

```text
ImportRequest(REPLACE, "s", ("K.1", MuseumNumber.of("K.1"))).fragment_ids  -> ('K.1', MuseumNumber(prefix='K', number='1', suffix=''))
ImportRequest(REPLACE, "s", "K.1").fragment_ids                            -> ('K', '.', '1')
```

The first is two kinds of id in one id list — exactly what the data-shape rule forbids — and it also defeats the de-duplication you added for my round-2 Finding 4 (`"K.1"` and `MuseumNumber.of("K.1")` are the same fragment but both survive). The second quietly turns one id into three. `5788e442` already added `deep_iterable(instance_of(...))` to `Media.projects` and `Media.references`; the same validator on `fragment_ids` (and a guard against a bare `str`) closes it. Please add the two cases above as tests.

#### L1 — Low — report and request contracts don't quite round-trip

`ebl/media/application/media_requests.py:95-101, 117-120, 138-144`.

```text
BackfillRequest(resume_after=BackfillReport(next_resume_token="").next_resume_token)  -> ValueError: resume_after cannot be blank
BackfillReport(scanned=-1).scanned                                                    -> -1
ImportReport(errors="oops").errors                                                    -> ('o', 'o', 'p', 's')
BackfillRequest(batch_size=10**12).batch_size                                         -> 1000000000000
```

The docstring says to pass `next_resume_token` back as `resume_after`, but a report can legally hold a token the request rejects. Counters accept negative values, `errors` / `warnings` split a bare string into characters (the `_strings_of` converter predates my last round), and `batch_size` has no upper bound despite `BackfillRequest` being described as "one bounded backfill batch". Suggested: validate the token like `resume_after`, `non_negative_int` on the counters, reject a bare `str` in `_strings_of`, and a maximum `batch_size`.

#### L2 — Low — invalid input raises three different exception types

`ebl/media/domain/representations.py:37`, `ebl/media/domain/media.py:45-53, 150-164`.

`MediaRepresentation(None, 1, 1, 1)` raises `AttributeError` (the mime converter runs before the validator); an association that isn't a `MediaAssociation` raises `AttributeError` in `_validate_associations` (no type check before `.fragment_id`); the new `projects` / `references` validators raise `TypeError`; every other invariant raises `ValueError`. `caption` and `import_source` aren't validated at all. Harmless while nothing calls these, but once routes or importers exist only `DataError` becomes a 422 — `AttributeError` / `TypeError` become 500s. Worth making them uniform now, while the domain is the only caller.

#### L3 — Low — percent-encoding does not make slash museum numbers routable

`ebl/media/application/media_urls.py:14-21, 30-31`; pinned by `ebl/tests/media/test_media_urls.py:20, 36, 50`.

The docstring contrasts the encoded media builders with the legacy path, implying encoding avoids the "extra path segment" problem for a museum number such as `A/B.1`. With Falcon 3.1.3 it doesn't — the path is decoded before routing:

```text
route /fragments/{number}/media/{media_id}/file
GET /fragments/A%2FB.1/media/x/file  -> 404
GET /fragments/K%201/media/x/file    -> 200, number == 'K 1'
```

So the tests pin URLs the future routes will not be able to serve. Nothing is broken today, but the docstring should say so plainly, so whoever registers the routes doesn't rely on it — the real fix (a different number encoding, or a query parameter) belongs with the routes.

#### L4 — Low — the PR description still documents the deleted docs

The "Documentation" section of the description describes `docs/media-architecture.md` and `docs/media-importer-backfill-security.md` ("Document the media architecture foundation…", "Document future media importer, backfill…"). Both were removed in `8eab5e6f`. Please update it to say where that material lives now (the module docstrings), so the description matches what merges.

#### I1 — Info — two backfill-request tests are duplicated across files

`ebl/tests/media/test_media_representation_store_contracts.py:176-181` repeats `test_media_request_validation.py:42-47` (defaults), and `:184-192` repeats `:76-83` (bounded resumed run) with only the cursor string changed. The backfill tests also sit in a file named for representation-store contracts. Keep the ones in `test_media_request_validation.py` and drop the copies — this is a test *move* of duplicates, but per our rules I'd want it stated in the PR rather than done silently.

#### I2 — Info — a contract test only exercises the in-memory fake

`ebl/tests/media/test_media_service_contracts.py:91-100`, `test_batch_fragment_read_does_not_delegate_to_per_fragment_reads`, subclasses `InMemoryMediaRepository` and asserts on it. It pins the fake's behaviour, not a contract any production class has to meet. Fine to keep as documentation of intent, but it shouldn't be counted as coverage of the batch-read requirement.

#### I3 — Info — carried over from round 2

- `fragmentarium` → `media` dependency edge (`fragment_query_summary_schema.py:20-21`) — unchanged; accepted.
- Route-drift test between Falcon templates and `media_urls` builders — still deferred to when routes land (`media_urls.py:9-12`); accepted.
- `FragmentMediaResponseDto.media` holds PHOTO and COPY media in one array told apart by `type`. They are the same data type (`Media`) with a category field and identical shape, so I don't read it as a data-shape violation — noting it only so it stays a deliberate choice when the routes are designed.

#### I4 — Info — two pre-existing uncovered lines in the touched schema file

`ebl/fragmentarium/application/fragment_query_summary_schema.py:72, 168` are uncovered (98% on the file). Both come from `ece18e14` on master and this PR doesn't touch them — it changes only lines 20-21 and 153-154 — so the touched-lines rule is met. #751 rewrites this file and covers it at 100%, so it resolves itself when both land.

#### I5 — Info — a merge commit carries 2,000 lines of its own changes

`fb6a6d57` ("fix: address more bugs") is a merge of `a106548a` that also deletes `media.py` and creates `media_repository.py`, `media_service.py`, `media_store.py`, `media_stored.py`, `representations.py` and `validation.py`. It predates my round 2, so that round saw its result; flagging it only because a code change inside a merge is invisible to `git log -p` and bisect. Please keep merges pure going forward.

PR #740, opened 2026-07-16
Realia annotation API: resolve realiaInfo on every fragment-returning route

khoidt:
## Summary

Adds the realia annotation API and fixes a defect where `realiaInfo` was resolved on **only one** of the 13 routes that return a fragment DTO.

`GET /fragments/{number}` was the sole `create_response_dto` call site passing `realia_info`. Every write route omitted it, so the frontend — which feeds these responses straight back into its fragment state — degraded realia labels from their lemma to the raw `realia_000846` id after **any** edit, not just after saving annotations.

Note `FragmentSchema.filter_none` (`@post_dump` → `pydash.omit_by(data, pydash.is_none)`) stripped the `None`, so the key was **absent**, not `null`. Both are invisible to a client that treats the field as optional, which is exactly why this went unnoticed.

## Approach

`realia_info` is now a **required** argument of `create_response_dto`, so omitting it is a type error rather than a silent failure. A new `FragmentDtoFactory` owns resolution in one place and is injected into every fragment-returning resource, so no route can forget it and the resolve call isn't duplicated 13 times.

`realiaInfo` is now always present on every fragment-returning route: `[]` or the resolved list.

### Error semantics for the realia lookup

A realia-store failure is handled **asymmetrically, by design**:

- **Read path — degrades.** `_find_by_realia_ids` catches `PyMongoError`, logs a warning, and returns `[]`, so the response is a `200` with `realiaInfo: []`. This closes the post-write failure window raised in review: the fragment update is already committed by the time the DTO is built, so raising there would return a `500` for a write that actually succeeded and invite an unsafe retry.
- **Validation path — fails hard.** `NamedEntityResource._validate_realia_ids` calls `find_by_realia_ids` **directly**, with no `PyMongoError` handling. That call happens *before* the write, so failing there is safe and correct: nothing has been persisted, and silently accepting an unverified `realiaId` would be worse than a `500`.

Please keep this split if you touch that code — it is deliberate, not an oversight.

The known trade-off is that a degraded `realiaInfo: []` is indistinguishable from a fragment that genuinely has no realia, while the `realia` array still lists the annotations. The client will render raw ids until the store recovers. The only signal is the server-side warning.

## Annotations across a transliteration update

Annotations now follow the rule: **if the transliteration a word carries changed, its annotations reset; if it did not, they are kept.**

`AbstractWord` merging already preserved annotations on words whose value was unchanged, but a fragment-level entry was left behind when its word *did* change — so `GET /named-entities` returned that entry with an empty `span`. `Fragment.drop_orphaned_annotations` now prunes `named_entities` and `realia` to those still referenced by at least one word. Covered by `ebl/tests/fragmentarium/test_annotations_after_transliteration.py`.

## Also in this PR

- **Missing test filled.** `ColophonResource.on_post` had *no test at all*; only the names GET was covered. A line this change touches was uncovered, so `test_update_colophon` was added. `colophons.py` goes 82% → 100%.
- **Type-checker gate gap closed.** This repo runs three type checkers. `task type` (pyre) and mypy both pass, but **pyright** — what Pylance shows in the IDE — was in no gate, so pyright errors could survive a green build. Adds a diff-scoped `task type-pyright` (pinned `pyright@1.1.411`, mirroring the pinned-`npx` precedent in `lint-md`), wired into `test-all`. Diff-scoped because the tree carries pre-existing errors; this enforces that *new* code stays clean.
- **`update_field("ocredSigns", …)` was broken.** The mapping was `("ocredSigns")` — a string, not a tuple — so `Schema(only=…)` raised `StringNotCollectionError` on every call. Now `("ocred_signs",)`, using the attribute name `only=` expects rather than the `data_key`, and covered by `test_update_ocred_signs`.
- **Devcontainer port fix.** `task start` binds **8001** but only **8000** was forwarded, so the browser could not reach the API at `localhost:8001` and DevTools reported "Failed to load response data". Now forwards both, labelled by how the API is started. **Requires a devcontainer rebuild**; forward 8001 from the Ports panel for an immediate unblock.

## Verification

Driven against a running instance of the modified service: `GET /fragments/{number}` returns `namedEntities`, `realia`, `realiaInfo` and per-word id arrays; `/named-entities` returns the object form with spans; `POST /genres` and `POST /edition` — plain write routes — now carry `realiaInfo`, which is the defect being fixed.

Negative cases confirmed live: `realiaId` inside a `namedEntities` element is a **422** (`Unknown field.`), an unknown `realiaId` is a **422**, and the same `id` in both arrays is a **422**.

- Full suite: **0 failures**
- Coverage: **100%** on all changed modules
- pyre, pyright, mypy, ruff format/lint, markdownlint: all clean

## Notes for the reviewer

- The `namedEntities` / `realia` split is preserved end to end — separate arrays at domain, Mongo, and wire level, no discriminator, and uniqueness is still enforced across the **union** of both (`_validate_unique_ids`).
- Because each array holds exactly one type, marshmallow's default `unknown=RAISE` rejects a mixed payload for free — no discriminator, no `OneOfSchema`, no bespoke validator.
- `edition.py:58` (`except ValueError` in `_create_transliteration`) remains uncovered. Its content is untouched by this PR, and it looks unreachable: the factory raises `TransliterationError`, not `ValueError`. Possible dead code, flagged rather than fixed.
- **Overlap with #743.** #743 ("Make the ATF parser visible to the type checkers") was split out of this branch and touches `ebl/tests/factories/archaeology.py` and `ebl/tests/fragmentarium/test_fragment_repository_updates.py` too. The `WordId` hunk is identical in both. This branch takes `archaeology.py` further — #743 fixes only the `reportPrivateImportUsage` errors, while this branch also removes the last 9 pyright errors by moving the factories to `make_factory`, which builds `Meta` internally. Expect a conflict there and prefer this branch's version.
- The one remaining mypy error on this branch (`fragment_metadata.py` — `lark_parser` "has no attribute `parse_markup_paragraphs`") is caused by the `lark_parser.py` / `lark_parser/` module-vs-package collision and is fixed by **#743**'s grammar-directory rename, deliberately not duplicated here.
- **qlty comments.** The 10 `function-parameters` findings are all pytest fixture parameters; bundling fixtures into a container purely to satisfy a counter would make the tests worse, so they are acknowledged rather than actioned. The 2 `similar-code` findings in `token_schemas_words.py` **are fixed** — the three duplicated `make_token` bodies now share `shared_word_arguments`/`load_word`, which also removed the `word_class` `ClassVar` that was failing pyre.
- Sourcery declined to review this PR ("larger than the review limit of 150000 diff characters"), so there is no automated logic review on the change.
- **All `TASK-740-*.md` tracking files must be removed before merge.**


Fabdulla1 (changes requested):
Overall, everything seems fine. I only have a couple of concerns. In `ebl/fragmentarium/web/dtos.py`, the mutation is persisted before `FragmentDtoFactory.create()` performs additional Realia lookup. If the lookup fails, I believe the client could receive a 500 even though the update has already been committed. I noticed this would leave the result of the request ambiguous and encourages a retry. Maybe we could avoid the post-write failure window, reuse already resolved data where available, or explicitly test and document the retry semantics? 

One other concern I had was that tests cover missing Realia records being omitted but I couldn't find coverage for an actual infrastructure failure from `find_by_realia_ids` such as a Mongo timeout or connection error. Since the branch allows those failures to propagate it might be worth it to add a test confirming the resulting HTTP behavior? 

Thanks for the hard work! :D

Fabdulla1 (approved):
Fantastic work B-) There are is only one small change that you'd just need to change before merging. Honestly, just strengthen the no-retry regression test in `ebl/tests/fragmentarium/test_realia_info_route.py:198-231`. It would be great if it also could assert that the update operation was performed exactly once and that the expected validation-plus-response lookup count was not exceeded. I think that would protect the original concern that a successful write musn't become ambiguous or encourage an unsafe retry. After that, feel free to merge :D

PR #741, opened 2026-07-20
Fix AfO Register texts-numbers match for references containing spaces

khoidt:
## Problem

`POST /afo-register/texts-numbers` returned `[]` whenever a reference's
`text` or `textNumber` contained a space (e.g. `"OrNS 59, 17"`), breaking
the Fragment → AfO Register link — fragments showed no AfO entry even
though the AfO Register correctly listed the fragment
(<https://www.ebl.lmu.de/library/MNAO.11676>).

## Root cause

The endpoint recovered `(text, textNumber)` from the joined string by
splitting on the last space. Both fields routinely contain spaces, so this
is unrecoverable in principle: `"OrNS 59, 17"` split into `text="OrNS 59,"`,
`textNumber="17"`, matching no document. The concat pipeline that would have
handled it was only reached when a split *failed*, which never happened for
the broken inputs.

## Fix

Replace the split with `candidate_splits`, which enumerates every possible
split point and `$or`s them — provably equivalent to matching the
concatenation while still using the compound `(text, textNumber)` index.
Removed `split_text_and_number`, `_build_indexed_query`, and the
now-unreachable `_build_fallback_pipeline`. Endpoint contract unchanged
(array of strings in, records out); no frontend change needed.

## Tests

Regression coverage at repository and route level: space in `textNumber`,
spaces in both fields, partial-match rejection, batched spaced/unspaced
queries, empty and space-free queries. Existing tests pass unchanged. Also
filled a pre-existing coverage gap in `create_search_query` in the touched
file.

## Verification

- Full suite: 3848 passed, 0 failures; 100% coverage on the changed file.
- Ran the modified service against a local mongo seeded with MNAO.11676's
  real shapes; all cases correct, including no false positive against a
  `"OrNS"/"59, 170"` decoy.
- Not verified here (needs deploy): MNAO.11676 rendering its AfO entries.

## Note to reviewers — branch history

While opening this PR, a push accidentally fast-forwarded `master` directly
to the fix commit. It was reverted immediately (`5967652a`), so **master's
tree is back to its pre-change state** — but master's *history* now carries
the original fix commit plus its revert. This branch re-applies the fix as a
fresh commit (`e43777ba`) from the original base, so merging this PR restores
the intended behaviour cleanly. No action needed from reviewers on that.

## Before merge

`TASK-741-{todo,log,review}.md` and `TASK-741-fix-{todo,log}.md` are
task-tracking scaffolding and should be removed before merge.


Fabdulla1 (changes requested):
in `ebl/afo_register/web/afo_register_records.py` the number of generated `$or` clauses grow with both the number of submitted references and the number of whitespace-separated tokens in each reference. Since the endpoint is publicly accessible with no request size or candidate-count validation, I think it might be worth adding a reasonable upper bound on the array length, individual string length or even the total generated candidates. I think this could protect from unexpected large MongoDB queries

For `ebl/tests/afo_register/test_afo_register_repository.py`, would you be able to add coverage for an ambiguous joined reference where two different records both match the same input? For example, ("A", "B C") and ("A B", "C") both concatenate to "A B C". I think the new candidate split approach should presumably return both records so an explicit test would document and protect the intended multi-match behavior.

In `ebl/afo_register/infrastructure/mongo_afo_register_repository.py` duplicate request strings or overlapping candidate pairs can currently produce repeated identical `$or` clauses. I think MongoDB should still return correct records, but deduplicating candidates by (text, textNumber) before constructing the query would keep the query smaller and avoid unnecessary planner work.

ebl/afo_register/web/afo_register_records.py
```diff
@@ -1,12 +1,41 @@
...
+MAX_QUERY_TOKENS = 24
+
+
+def validate_query(query: object) -> str:
+    if not isinstance(query, str):
+        raise DataError("Each query must be a string.")
+    if len(query) > MAX_QUERY_LENGTH:
+        raise DataError(
+            f"Query too long: at most {MAX_QUERY_LENGTH} characters allowed."
+        )
```
Fabdulla1: So `MAX_QUERY_LENGTH` here is enforced with `len(query)` which counts python code points rather than the UTF-8 bytes. The concern I see here is that a request can satisfy every new bound while still producing a MongoDB filter larger than the 16 MB BSON limit. 

For example, with 4-byte UTF-8 chars, 500 chars / 24 token strings can reach the 10,000 candidate limit while producing an `$or` filter of roughly 18.6 MB. That would recreate the oversied-query this is trying to eliminate just through byte width rather than candidate count. 

I think it would be good to bound the UTF-8 encoded length instead. Maybe `len(query.encode("utf-8"))` or enforce a safe BSON-size bound before `find_many`?

Fabdulla1 (approved):
Overall great! Just address the comment above and it should be good to merge!

PR #749, opened 2026-08-06
Add new manuscript types

ejimsan:
## Summary by Sourcery

Add support for additional manuscript types across the domain model, parser, tests, and ATF documentation.

New Features:
- Introduce new manuscript types (multi-column, collective, student-teacher, school lentil, prism, uncertain) to the ManuscriptType enum and ATF abbreviations.

Documentation:
- Document the new manuscript type abbreviations in the EBL ATF specification.

Tests:
- Extend chapter parsing tests to cover the newly added manuscript types.


Fabdulla1 (changes requested):
For the backend, the only comment I have is that the `Unc` period/type adjacency would need to be verified. `Unc` is now accepted as a `TYPE` while it already exists as both a `PERIOD` and a `PROVENANCE` abbreviation. The grammar structure makes this look likely to be safe and existing tests exercise `TYPE=Unc` after ordinary periods, but I don't see a case where `Period.UNCERTAIN` is immediately followed by `ManuscriptType.UNCERTAIN`. It would be good to add or confirm a focused parser case for that exact combination so we know the new triple use of `Unc` is unambiguous.

Fabdulla1 (approved):

PR #751, opened 2026-08-14
dev: add support for frontend fix-n-calls

Fabdulla1:
This branch enriches the bounded `GET /fragments/query` response so frontend #787 can render result cards without per-card fragment, thumbnail, or bibliography requests. The enrichment applies only when `limit` is present; the existing no-limit `QueryResult` path keeps its previous response contract.

## Limited query response

The limited response includes summary metadata, complete `matchingLines` and `matchCount`, and up to five unique matching `TextLine` previews. Each preview line is the same rich TextLine representation used by fragment detail responses, plus its source `index`. Five is a line-count cap, not a byte-size guarantee.

`bibliographyDocuments` maps each original reference ID used by the returned page to its successfully resolved, active canonical CSL-JSON document. The key remains the ID on the fragment reference even when the value's canonical `id` differs. Missing initial entries, dangling redirects, cycles, chains requiring more than five redirects, and deprecated entries without a usable target are omitted independently; other references on the page remain available. A live record reached in exactly five redirects is included. Resolution uses one deduplicated batch per redirect hop, is bounded to the returned page, and performs no bibliography query when that page has no references. Database or other operational failures still fail the request instead of being treated as missing bibliography data.

## Query limit and pagination

- Nonpositive or non-integer `limit` values return `422`.
- Values above 200 are clamped to 200.
- Values from 1 through 200 are accepted unchanged.

Clients must paginate using `hasNextPage` and the existing offset/count semantics. Frontend #787 currently requests page size plus one (26, 51, or 101), all below the backend maximum, and consumes both overfetch and `hasNextPage` signals.

## Coordination

This backend change is intended to release with ElectronicBabylonianLiterature/ebl-frontend#787. The frontend treats a bibliography ID absent from the map as an unresolved citation while preserving valid sibling citations and the rest of the card. The frontend PR remains a separate, read-only dependency for this backend change. `/fragments/latest` enrichment is outside this PR.


khoidt (changes requested):
Nice piece of work — this does what it says on the tin. The N+1 is gone, the bibliography lookup is properly batched and deduplicated, and the redirect handling that Codex flagged has been fixed well (multi-hop now resolves, one batch per hop, original key retained). The test suite around it is genuinely thorough — the redirect edge cases, page-bounding, and the "no references means no query" short-circuit are all covered. Full suite is green, every line the PR adds is covered, and I confirmed all of it against a live server, not just tests.

Two things I'd want sorted before merge. First, `task type-pyright` is red: 38 new errors, all in the new `test_fragment_query_preview.py`, all the same fix that the production code already uses (`cast(dict, ...)`). That's a hard gate, so it blocks. Second, `/fragments/query` will happily hand the frontend a *deprecated* bibliography record in the cases where `/bibliography/{id}` returns 404 — I reproduced both live. That's probably a deliberate "don't fail the whole search" choice, but the two endpoints currently disagree and nothing says so.

Beyond that, a few things worth a conversation rather than a blocker: preview lines now ship `content` *and* a compact `tokens`/`text` copy of the same data (3156 vs 241 bytes per line — one of them is redundant), the `limit` bound is a breaking change for any client sending more than 200, and `parserVersion` moving out of `data_key` pushes wire-format naming into the domain object. Details below.

### Details

#### F1 — `task type-pyright` fails: 38 new errors (blocking)

**Severity: High (hard-gate failure)** · `ebl/tests/fragmentarium/test_fragment_query_preview.py`

Gate 4 of the pre-commit checklist is `task type-pyright` with **zero errors**. It currently reports **80 errors, 0 warnings**.

I ran a baseline in a detached worktree at the merge base `c2b0a5ef` to split them:

| File | Errors at HEAD | Errors at merge base | New |
| --- | --- | --- | --- |
| `ebl/tests/fragmentarium/test_fragment_query_preview.py` | 38 | *(file did not exist)* | **38** |
| `ebl/tests/fragmentarium/test_fragment_query_summary_schema.py` | 25 | 25 | 0 |
| `ebl/fragmentarium/infrastructure/mongo_fragment_repository_get.py` | 11 | 11 | 0 |
| `ebl/tests/fragmentarium/fragment_query_test_helpers.py` | 5 | 5 | 0 |
| `ebl/tests/fragmentarium/test_fragments_search_route_transliteration.py` | 1 | 1 | 0 |
| **Total** | **80** | **42** | **38** |

All 38 new ones are the same root cause: `OneOfLineSchema().dump(line)` has a broad union return type in the marshmallow stubs, and the test subscripts it directly. Breakdown: 24 `reportArgumentType`, 9 `reportCallIssue`, 5 `reportOptionalSubscript`, at lines 28, 30–33, 37, 52, 64, 73, 81, 89, 105, 113, 125, 201, 205, 218, 225, 228.

The production code already solves exactly this, at [fragment_query_preview.py:63](ebl/fragmentarium/application/fragment_query_preview.py#L63):

```python
preview_line_of(cast(dict, schema.dump(line)))
```

The test file just needs the same narrowing — a one-line helper used throughout:

```python
def dumped(line) -> dict:
    return cast(dict, OneOfLineSchema().dump(line))
```

Note the 42 pre-existing errors are not this PR's doing, but they are in files this PR touches, so they're worth a follow-up ticket.

#### F2 — `/fragments/query` serves deprecated bibliography records that `/bibliography/{id}` refuses

**Severity: Medium** · [`ebl/fragmentarium/application/fragment_query_bibliography.py:31-38`](ebl/fragmentarium/application/fragment_query_bibliography.py#L31-L38)

`resolved_document` stops walking and returns whatever it currently holds in three situations:

1. the redirect target does not exist (dangling `redirectTo`),
2. the chain forms a cycle,
3. the chain is longer than `MAX_REDIRECT_DEPTH`.

In all three, the returned document is still `deprecated: true`. `Bibliography._follow_redirect` — the canonical resolver at [`bibliography.py:59-91`](ebl/bibliography/application/bibliography.py#L59-L91) — *raises* in exactly those three cases (`NotFoundError`, `DuplicateError`, `DuplicateError`).

Reproduced against a running server (see Reproduction Steps for setup):

```text
/bibliography/DANGLE751                          -> 404
/fragments/query?...&limit=10                    -> 200
  bibliographyDocuments["DANGLE751"] = { "id": "DANGLE751", "deprecated": true, "redirectTo": "NOPE751" }
```

So the same entity is a 404 on one endpoint and a silently-deprecated 200 payload on the other. The behaviour is deliberate enough to be pinned by three tests — `test_dangling_redirect_keeps_the_stored_record`, `test_redirect_cycle_terminates_without_extra_queries`, `test_chain_longer_than_the_redirect_depth_stops_safely` — but it is not written down anywhere, and a frontend that renders `bibliographyDocuments[id]` without checking `deprecated` will show a deprecated citation.

Worth noting the inconsistency inside this module too: a reference id with **no** stored entry is omitted from the map entirely (`test_query_omits_missing_bibliography_documents`), but a reference id whose *target* is missing gets a deprecated record instead. Two flavours of "unresolvable", two different answers.

**Recommendation:** pick one contract and state it. Either omit unresolvable entries from the map (consistent with missing ids — the client falls back to fetching), or keep them and document in the API contract that a returned document may carry `deprecated: true`.

#### F3 — Preview lines carry the same data twice: `content` and `tokens`/`text`

**Severity: Medium** · [`fragment_query_preview.py:25-38`](ebl/fragmentarium/application/fragment_query_preview.py#L25-L38), [`fragment_query_summary_schema.py:97-107`](ebl/fragmentarium/application/fragment_query_summary_schema.py#L97-L107)

`preview_line_of` emits both a full `content` (the untouched `OneOfLineSchema` token list) and a compact projection (`tokens`, `text`). Their own test asserts the duplication is exact — [test_fragment_query_preview.py:33](ebl/tests/fragmentarium/test_fragment_query_preview.py#L33):

```python
assert preview["content"] == detail["content"]
```

Measured on one real preview line from the running service:

| Field | Bytes |
| --- | --- |
| `tokens` | 241 |
| `content` | 3156 |
| whole preview line | 3632 |

Every field in `tokens[i]` (`value`, `cleanValue`, `uniqueLemma`, `type`) is also present in `content[i]`, and `text` is just `" ".join(token["value"] for token in content)`. So `tokens` and `text` are 100% derivable client-side.

The PR description says it "caps preview payload size". The *line count* is capped at 5, but each line grew roughly 15× versus the old compact-only shape, and ~7% of what ships is a duplicate of what's next to it. More importantly, two representations of the same thing with no stated authority is a maintenance trap — a future change to token serialization has to remember both.

**Recommendation:** drop `tokens` and `text` from both `preview_line_of` and `FragmentQueryPreviewLineSchema` and let the frontend derive them from `content`; or, if the frontend genuinely wants the compact form, drop `content`. One source of truth either way.

#### F4 — `limit` is now bounded to 1–200; this is a breaking API change

**Severity: Medium** · [`parameter_parser.py:41-45`](ebl/common/query/parameter_parser.py#L41-L45)

`limit` was previously unbounded. Verified live on this branch:

```text
limit=0    -> 422 "limit must be between 1 and 200"
limit=1    -> 200
limit=200  -> 200
limit=201  -> 422
limit=-1   -> 422
```

Any existing caller sending `limit=500`, `limit=1000` or `limit=0` gets a 422 where it used to get a 200. That is a reasonable hardening, and nothing inside this repo passes a large limit to `/fragments/query` (I checked — `RETRIEVE_ALL_LIMIT = 1000` is an internal `$limit` stage, not the query param, and no corpus route accepts a user-supplied `limit`). But the PR body describes it as "bounds query limits" without flagging it as breaking.

Also a consistency point: the bibliography endpoints solve the same problem by **clamping** rather than rejecting — [`bibliography_entries.py:109`](ebl/bibliography/web/bibliography_entries.py#L109) uses `min(req.get_param_as_int("limit") or 10, 25)` and [`:119`](ebl/bibliography/web/bibliography_entries.py#L119) uses `min(..., 100)`. Two conventions for the same concern now live in the codebase.

**Recommendation:** confirm the frontend never sends more than 200, and call the change out explicitly in the PR description / release notes. Optionally align on one convention (clamp vs 422) across the API.

#### F5 — `parserVersion` moved out of `data_key`, pushing wire naming into the domain

**Severity: Medium-Low** · [`fragment_query_summary_schema.py:115`](ebl/fragmentarium/application/fragment_query_summary_schema.py#L115), [`fragment_query_summary.py:21-22`](ebl/fragmentarium/domain/fragment_query_summary.py#L21-L22)

The change:

```python
-    parser_version = fields.String(required=True, data_key="parserVersion")
+    parserVersion = fields.String(required=True)
```

with `empty_matching_line_preview()` following suit (`{"lines": (), "parserVersion": ...}`).

Every other field in this file uses snake_case attributes with an explicit `data_key`: `matching_lines`/`matchingLines`, `match_count`/`matchCount`, `has_photo`/`hasPhoto`, `thumbnail_path`/`thumbnailPath`, `bibliography_documents`/`bibliographyDocuments` — including the one this PR adds. The new `lineNumber` field breaks the convention the same way.

The practical effect: `FragmentQuerySummary.matching_line_preview` is a `Dict[str, Any]`, and it now holds camelCase wire keys inside the domain object.

This *does* fix Sourcery's overall comment (the two builders no longer disagree — both emit `parserVersion`), but it fixes it in the direction that leaks the wire contract inward. The alternative closes the same gap without that cost: keep `data_key="parserVersion"` with a `parser_version` attribute, and have `matching_line_preview_of_data` emit `parser_version`.

#### F6 — `matchingLines` and `matchingLinePreview.lines` are no longer positionally related

**Severity: Low** · [`fragment_query_preview.py:40-45`](ebl/fragmentarium/application/fragment_query_preview.py#L40-L45)

`selected_lines` deduplicates and caps at `MAX_PREVIEW_LINES = 5`; `matchingLines` on the summary keeps every index, duplicates included. Live, on a 12-line fragment:

```text
single-line query:  matchCount=12   len(matchingLines)=12   len(preview.lines)=5
two-line query:     matchingLines = [0,1,1,2,2,3,3,4,4,5,5,6,6,7,7,8,8,9,9,10,10,11]
                    preview.lines numbers = ['1.', '2.', '3.', '4.', '5.']
```

A client that zips `matchingLines[i]` with `preview.lines[i]` gets wrong data. It *can* recover the relation by string-matching `number`/`prefix`, but nothing in the response makes it explicit.

**Recommendation:** carry the source index on each preview line (`{"index": 3, ...}`) so the link to `matchingLines` is stated rather than inferred.

#### F7 — Data hard gate: the preview line schema gained per-type optional fields

**Severity: Low** · [`fragment_query_summary_schema.py:97-107`](ebl/fragmentarium/application/fragment_query_summary_schema.py#L97-L107)

Before this PR, `FragmentQueryPreviewLineSchema` had four always-present fields (`number`, `prefix`, `text`, `tokens`). The PR adds `type`, `lineNumber` and `content`, and `preview_line_of` filters `None` values out — so `lineNumber` and `content` are present for text lines and absent for line types that lack them, with `type` acting as the discriminator a reader has to probe.

That is the pattern the data hard gate names directly: *"An optional field present for one type and absent for another means you have two types in one array."*

To be fair about the severity: I could not produce a non-`TextLine` preview line against the running service. `selected_lines` only ever picks indices that came out of a sign match, so in practice the array holds one type today. The shape *permits* more, and this is inherited from `OneOfLineSchema`, which is the repo-wide representation of a transliteration line and far outside this PR's scope to re-model.

**Recommendation:** make the decision explicit rather than incidental — either narrow the schema to text lines (`lineNumber` and `content` `required=True`, and type the helper as taking a dumped text line), or note in the module that the preview intentionally mirrors `OneOfLineSchema` polymorphism.

#### F8 — `mongo_fragment_repository_get.py` is 500 lines, 2× the 250-line hard gate

**Severity: Low** · `ebl/fragmentarium/infrastructure/mongo_fragment_repository_get.py`

Pre-existing (535 lines at the merge base). This PR *improves* it — extracting `compact_preview_token`, `compact_preview_line` and `_matching_line_preview` into `fragment_query_preview.py` brings it to 500. It is still double the limit, and this PR is already the one doing the extracting, so it is the natural moment to finish the job.

Every other changed file is comfortably under 250 (largest: `test_fragments_search_route_filters.py` at 247, `test_fragment_query_preview.py` at 238).

#### F9 — mypy: 2 errors in changed files, both pre-existing false positives

**Severity: Informational** · gate 8 only

`poetry run mypy <changed> --ignore-missing-imports` reports 52 errors, but only 2 are located in the 22 changed files:

```text
ebl/tests/fragmentarium/test_fragment_query_preview.py:12: error: Module "ebl.transliteration.domain.atf_parsers.lark_parser" has no attribute "parse_atf_lark"  [attr-defined]
ebl/tests/fragmentarium/test_fragments_search_route_preview.py:8: error: (same)
```

The import is correct and identical to the ~10 other places in the codebase that use it. Verified pre-existing: the same error fires on untouched `ebl/tests/fragmentarium/test_fragment.py:36`. Not a defect in this PR — flagged only because gate 8 reads red.

#### F10 — Every limited query now costs 1–6 extra sequential round trips

**Severity: Low** · [`fragments.py:139-146`](ebl/fragmentarium/web/fragments.py#L139-L146), [`fragment_query_bibliography.py:50-66`](ebl/fragmentarium/application/fragment_query_bibliography.py#L50-L66)

`_enrich` runs on every `limit` query. The design is good — one batched `$in` per hop, ids deduplicated in first-seen order via `dict.fromkeys`, and a clean short-circuit to zero queries when the page has no references (confirmed live and by `test_query_without_references_skips_bibliography_lookup`). Typical cost is exactly one extra query; the redirect batches only fire when deprecated entries are on the page.

Two things to keep an eye on:

- The hops are sequential, so a page containing a 5-deep chain costs 6 serial round trips on the hot search path.
- A canonical document is repeated once per aliasing id. Live, `CANON751` appears under both `"CANON751"` and `"OLD751"` — two full CSL-JSON copies. At `limit=200` with heavily-referenced fragments this adds up.

**Recommendation:** no change needed now. If payload becomes a concern, key `bibliographyDocuments` by canonical id and add a small `referenceId -> canonicalId` map alongside it.

khoidt (changes requested):
This has come a long way and most of it is genuinely good now. Every one of my ten findings has been dealt with, and not superficially — `content` is gone from the preview lines (a preview line is 338 bytes now, down from the 3632 I measured), the `index` field makes the link back to `matchingLines` explicit, `parserVersion` is back behind a `data_key` where it belongs, and `mongo_fragment_repository_get.py` went from 535 lines to 239 by being split into two proper modules. Codex's multi-hop redirect point is fixed too — I built a three-hop chain and it resolves to the final target, and a cycle terminates without hanging. Full suite green, 100% coverage on all ten changed modules, CI and qlty clean.

Two things stop me approving, though.

The blocking one is a new `.md` file — `docs/FRAGMENT_QUERY_BIBLIOGRAPHY_CONTRACT.md`. Awkwardly, it *is* the fix for my F2: I asked you to pick a contract and state it, and you stated it there. The content is right and worth keeping; it just cannot live in the repo. Move it to the PR description and the blocker goes away without losing anything.

The other is that `task type-pyright` is red again — 3 errors, gate 4, zero tolerance. Not the same 38 I found last time; those are properly fixed. These are 3 pre-existing errors in `test_fragment_repository_query.py` that were dormant because nothing had touched that file. The last commit appended one test to it, which pulled it into the gate's scope. Unlucky rather than careless, but the gate reads red either way.

Everything else below is small. Worth knowing that the last commit landed after Sourcery's approval, so nothing has reviewed it but me.

### Details

#### 1. [Blocking] A new `.md` file is added to the repository

`docs/FRAGMENT_QUERY_BIBLIOGRAPHY_CONTRACT.md` (66 lines), added by `48e8834e`. It is not referenced from any code or from any other document.

The awkward part is that this file is the fix for my F2, where I asked: *"pick one contract and state it. Either omit unresolvable entries from the map ... or keep them and document in the API contract that a returned document may carry `deprecated: true`."* The second option was chosen, and this file is where it was stated. The document is good — it is precise about all six resolution cases and explicit that a client must check the `deprecated` flag.

So deleting it alone would regress F2 back to undocumented. The fix has to be a move, not a deletion.

**Severity:** Blocking. No new markdown lands in this repository; that is not a judgement call.
**Reproduction:** `git diff --name-status <merge-base> HEAD -- '*.md'` → `A docs/FRAGMENT_QUERY_BIBLIOGRAPHY_CONTRACT.md`.
**Recommendation:** move the whole contract into the PR description, verbatim — it is short enough. If it needs to live in the code, the module docstring of `fragment_query_bibliography.py` is the natural home and is where a maintainer would look first. Either way, do not simply drop it: F2 stays answered.

#### 2. [High] `task type-pyright` fails — 3 errors, and the last commit is what exposed them

Gate 4 requires zero errors. Current state at `2d053f6f`:

```text
ebl/tests/fragmentarium/test_fragment_repository_query.py:237:15 - error: Object of type "() -> dict_items[Unknown, Unknown]" is not subscriptable (reportIndexIssue)
ebl/tests/fragmentarium/test_fragment_repository_query.py:237:22 - error: Cannot access attribute "items" for class "list[Unknown]" (reportAttributeAccessIssue)
ebl/tests/fragmentarium/test_fragment_repository_query.py:237:22 - error: "items" is not a known attribute of "None" (reportOptionalMemberAccess)
3 errors, 0 warnings, 0 informations
```

To be clear about what happened, because the credit matters: my F1 was 38 errors in `test_fragment_query_preview.py` and a total of 80 across the changed set. That file is now completely clean and the total is down to 3. The fix worked.

These 3 are different. They exist at the merge-base too — I ran pyright on that file in a baseline worktree and got the identical three errors on the identical line, which this PR never edits. They were dormant because `task type-pyright` only checks files changed against master, and that file was not in the changed set.

Then `2d053f6f` appended `test_query_fragmentarium_limit_summary_missing_museum_number_fails_clearly` to it. Changed-file count went 27 → 28, the file entered the gate's scope, and the gate went red:

```text
files changed at 48e8834e: 27   (pyright: 0 errors)
files changed at HEAD:     28   (pyright: 3 errors)
```

The repo rule is that a pre-existing error in a file you touch gets fixed rather than carried, and the same reasoning I applied to `bibliography_entries.py` on #752 applies here.

**Severity:** High — a hard-gate failure. CI does not run pyright, so nothing on GitHub catches it.
**Reproduction:** `task type-pyright` on this head.
**Recommendation:** narrow the `find_one` result at line 237 the way the production code already does. Either `assert ... is not None` before subscripting, or a `cast(dict, ...)`. It is a one-line fix and it clears the gate.

#### 3. [Low] `matching_line_preview_of` has no production caller

`ebl/fragmentarium/application/fragment_query_preview.py:60` is reached only from tests — `fragment_query_test_helpers.py`, `test_fragment_query_preview_selection.py`, `test_fragment_query_preview.py`, `test_fragment_query_summary_schema.py`. Production hydration goes through `matching_line_preview_of_data` from `mongo_fragment_repository_get_summary.py:122`.

Your own commit message for `2d053f6f` says as much — *"the domain-Text builder that has no production caller"* — and that commit moved the schema test off it for exactly that reason. So this is known; it just was not finished. It sits in `application/`, counts as production code, and reads at 100% coverage purely from tests, which makes it look exercised when it is not.

The two builders also emit deliberately different keys — `_of_data` emits `parserVersion` because its output is fed through schema *load*, `_of` emits `parser_version` because its output goes straight onto the domain object. That is correct for each layer and is the direction I asked for in F5, but nothing in the module says so, and the near-identical names invite someone to "fix" the inconsistency and break the load path.

**Severity:** Low.
**Recommendation:** either delete `matching_line_preview_of` and move its test coverage onto `_of_data`, or keep it and add one line to each function saying which layer's key shape it produces.

#### 4. [Low] `limit` above 200 is now silently clamped

F4 is resolved, and better than it first looked. I measured both sides rather than assuming:

| `limit` | master | this PR |
| --- | --- | --- |
| `0` | **500** | `422 "limit must be at least 1"` |
| `-1` | **500** | `422` |
| `201` | `200`, all results | `200`, capped at 200 |
| `1000` | `200`, all results | `200`, capped at 200 |

So the rejection path is a strict improvement — master answered `limit=0` with a 500. And large limits now clamp rather than 422, which is what I asked for and matches the `min(...)` convention the bibliography routes already use.

The one residual: a client sending `limit=1000` gets 200 rows and no signal that it was capped. `hasNextPage` covers a correct paginator, but a client that assumes it received everything it asked for will silently under-read.

**Severity:** Low. Worth one line in the PR description rather than a code change.

#### 5. [Low] The two "unresolvable" cases still answer differently

Documented now, but worth restating since it is the substance behind finding 1. A reference id with no stored entry is omitted from `bibliographyDocuments`; a reference id whose redirect target is missing gets a still-`deprecated` record. Reproduced live:

```text
GET /bibliography/DANGLE751                      -> 404
GET /fragments/query?number=X.1&limit=10
  bibliographyDocuments["GHOST751"]              -> absent        (no stored entry)
  bibliographyDocuments["DANGLE751"]             -> deprecated:true, redirectTo:"NOPE751"
  bibliographyDocuments["OLD751"]                -> id "CANON751" (resolved, 1 hop)
  bibliographyDocuments["HOP1"]                  -> id "FINAL751" (resolved, 3 hops)
```

Two flavours of unresolvable, two different answers, and the same id is a 404 on one endpoint and a 200 payload on the other. The new document states this plainly, which satisfies my F2 — the requirement was to state it, not to unify it. Recorded so that the statement survives wherever the document ends up.

#### 6. [Informational] mypy reads red on 3 changed files — all the same pre-existing false positive

```text
ebl/tests/fragmentarium/test_fragment_query_preview.py:12
ebl/tests/fragmentarium/test_fragment_query_preview_selection.py:14
ebl/tests/fragmentarium/test_fragments_search_route_preview.py:8
  error: Module "ebl.transliteration.domain.atf_parsers.lark_parser" has no attribute "parse_atf_lark"  [attr-defined]
```

The import is correct and used identically across the codebase. Confirmed pre-existing: the same error fires on untouched `ebl/fragmentarium/application/transliteration_update_factory.py:5` and `ebl/corpus/web/chapter_schemas.py:25`. Not a defect in this PR — flagged only because gate 8 reads red, exactly as I noted in F9.

#### 7. [Nit] `test_fragment_repository_query.py` is at 249 of the 250-line limit

The file the last commit appended to now sits one line under the hard gate. The next test added to it breaks a different gate. Worth splitting pre-emptively while you are in there for finding 2.

khoidt (changes requested):
This is very close. Every finding from my last round is dealt with: the contract doc moved into the description, the pyright errors are gone, the test-only preview builder moved into test helpers, and the nearly-full test file was split. I ran the whole thing end to end against a local database and the core of it works well — the `limit` bounds 422 where master used to 500, large limits clamp to 200, and a preview line is byte-for-byte the same line `GET /fragments/{n}` returns, with an `index` that points straight into `matchingLines`. Full suite green, 100% coverage on all ten changed modules, pyre and pyright clean.

What's left is mostly about the last commit. `b924d736` made over-depth chains disappear from `bibliographyDocuments`, but it decides that by counting hops, so a 5-member cycle vanishes while a 2-member cycle comes back flagged, and a chain that dangles at hop 5 vanishes while one that dangles at hop 1 comes back flagged. The description still documents the old behaviour. And the only client of this map — frontend #787, already approved — never looks at `deprecated`, so it will render a stale record as if it were current. Since the code is already omitting some unresolvable entries, the simplest fix is to omit all of them.

The one blocker is mechanical: mypy still reads red on three changed test files. It's the same `parse_atf_lark` false positive as before, but master has fixed it since (#743), and I checked that merging master clears it. So it's just "update the branch".

No new `.md` files, and no dev container, Docker, CI, dependency or config changes.

### Details

#### B1 — Blocker — mypy gate is red on three changed files; updating the branch clears it

`ebl/tests/fragmentarium/test_fragments_search_route_preview.py:8`, `test_fragment_query_preview_selection.py:14`, `test_fragment_query_preview.py:10`.

```text
error: Module "ebl.transliteration.domain.atf_parsers.lark_parser" has no attribute "parse_atf_lark"  [attr-defined]
```

This is the false positive I flagged as informational in both earlier rounds (F9 / #6), because at the time there was no fix. There is one now: master commit `cd46110c` ("Make the ATF parser visible to the type checkers", #743) landed after this branch's merge base. I built the merged tree (`git merge-tree origin/master <head>`, extracted without committing) and ran mypy on the three files: zero errors. Gate 8 requires zero errors in changed files, so this blocks until the branch is updated — nothing else to change.

#### M1 — Medium — redirect resolution now gives different answers to the same kind of unresolvable reference depending on hop count

`ebl/fragmentarium/application/fragment_query_bibliography.py:30-39`, introduced in `b924d736`.

```python
def resolved_document(document: dict, fetched: Dict[str, dict]) -> Optional[dict]:
    seen: Set[str] = set()
    while (target := redirect_target_of(document)) and target not in seen:
        seen.add(target)
        if target not in fetched:
            break
        document = fetched[target]
    if len(seen) >= MAX_REDIRECT_DEPTH and redirect_target_of(document):
        return None
    return document
```

`seen` counts cycle members and the unfetched dangling target as hops, so "is this over depth?" ends up depending on the cycle length or on where the chain breaks. Verified against the running service with one fragment citing every case:

| Reference | Shape | `bibliographyDocuments[id]` | `GET /bibliography/{id}` |
| --- | --- | --- | --- |
| `OLD` | 1 hop → live | `CANON`, canonical | — |
| `F0` | 5 hops → live | `F5`, canonical | 200 |
| `DANGLE` | dangling at hop 1 | itself, `deprecated: true` | 404 |
| `D0` | dangling at hop 5 | **omitted** | 404 |
| `A` | 2-cycle | itself, `deprecated: true` | 409 |
| `C5C` | 5-cycle | **omitted** | 409 |
| `H0` | 6 hops → live (over depth) | omitted | 409 |
| `NOTARGET` | deprecated, no `redirectTo` | itself, `deprecated: true` | 404 |
| `GHOST` | no entry | omitted | 404 |

So dangling-at-1 and dangling-at-5 disagree, and 2-cycle and 5-cycle disagree. The commit message for `b924d736` says *"a confirmed dangling target (query attempted, target doesn't exist) still resolves to the stored deprecated record"* — `D0` is exactly that case and it is omitted. The same two cases fail as unit tests:

```text
FAILED test_a_five_cycle_is_treated_like_a_two_cycle                     assert 'C' in documents
FAILED test_a_dangling_target_at_hop_five_is_treated_like_one_at_hop_one  KeyError: 'D0'
```

Fix: see M2 — omitting every unresolvable entry removes the hop counting entirely. If you'd rather keep flagged records, the omission test must be "the fetch loop ran out of hops while the next target was still unqueried", not `len(seen)`.

#### M2 — Medium — the documented contract is stale, and the only client doesn't honour it

PR description, "Bibliography documents" section; `fragment_query_summary_schema.py:163-168`; frontend `ElectronicBabylonianLiterature/ebl-frontend#787` at `cb78af15`.

In round 2 I accepted "keep unresolvable records and flag them `deprecated: true`" as a valid answer to F2, provided it was stated. Two things have moved since:

1. **The statement is now wrong.** The description still says a chain longer than `MAX_REDIRECT_DEPTH` returns *"the last document reached, still `deprecated: true`"*, but since `b924d736` it is omitted, and `test_chain_longer_than_the_redirect_depth_is_omitted` pins the omission. Add M1 and the real behaviour is three-way: resolved, omitted, or flagged, with the split between the last two decided by hop count.
2. **The consumer ignores the flag.** The description says *"A client that renders a citation from this map must check the `deprecated` flag."* #787 (approved) is that client, and its diff contains no reference to `deprecated` or `redirectTo` at all; `fragmentQueryMapping.ts` passes `dto.bibliographyDocuments` straight into `createQuerySummaryFragment`. A flagged record will be rendered as a current citation.

Recommendation: now that `b924d736` already omits some unresolvable entries, omit **all** of them — dangling, cyclic, over-depth, and deprecated-without-target. Then every value in `bibliographyDocuments` is a canonical document, the map holds one kind of thing, no client has to probe a flag, and M1 disappears. On the frontend side, #787 already has to handle ids missing from the map (`GHOST` today): `querySummaryFragment.ts:175` does `reference.document ?? bibliographyDocuments[reference.id]` and passes the result to `createReference`, inside a `try/catch` that degrades the card to `UnsupportedFragmentCardSummary`. Whether a whole card degrading because one citation is missing is the behaviour you want is worth confirming on #787 — I did not run the frontend. Update the description to match whichever way you go.

#### L1 — Low — the clamp route test passes without a clamp

`ebl/tests/fragmentarium/test_fragments_search_route_pagination.py:179-187`.

```python
@pytest.mark.parametrize("limit", [str(MAX_QUERY_LIMIT + 1), "1000000"])
def test_query_fragmentarium_limit_above_the_maximum_does_not_error(client, limit):
    result = client.simulate_get("/fragments/query", params={"number": "K.1", "limit": limit})
    assert result.status == falcon.HTTP_OK
```

It runs against an empty collection and only asserts 200, so it would pass with no clamp at all. The clamp itself is correct — `project=CAIC&limit=1000` over 206 matching fragments returns exactly 200 items at runtime — but only the parser unit test proves it. Seed `MAX_QUERY_LIMIT + 1` matching fragments and assert `len(items) == MAX_QUERY_LIMIT` (and `hasNextPage is True` with `count=page`).

#### L2 — Low — production `is_text_line` keeps a branch only the test helper uses

`ebl/fragmentarium/application/fragment_query_preview.py:9-11, 18-20`.

```python
def is_text_line(line: Any) -> bool:
    type_name = line.get("type") if isinstance(line, dict) else type(line).__name__
    return type_name == "TextLine"
```

Production always passes stored dicts; the domain-object branch exists only for `fragment_query_preview_test_helpers.matching_line_preview_of`. Consequences: the parameter is `Any`, `selected_lines` returns `List[Tuple[int, Any]]`, and most selection tests exercise the domain branch rather than the production dict path. I checked the dict path directly (indexes `(0, 1, 2)` over `1. / $ / 2.` → preview `[0, 2]`, it's right), but nothing in the suite pins "a matched `$` line is dropped" on the path production actually uses. Fix: have the test helper dump through `TextSchema` and call `matching_line_preview_of_data`, then narrow `is_text_line` to `dict`.

#### L3 — Low — preview lines are raw storage, not schema-validated

`ebl/fragmentarium/application/fragment_query_summary_schema.py:91` — `lines = fields.List(fields.Raw())`; filled at `mongo_fragment_repository_get_summary.py:122-124`.

Whatever keys are stored in Mongo become the wire format, so the backend schema is no longer what defines those field names. Today it's fine — at runtime a preview line equals `GET /fragments/{n}` → `text.lines[index]` exactly (checked on a seeded fragment). But that equality holds because two paths happen to agree, not because a schema enforces it. A nested schema based on the existing text-line schema plus `index` would make it a contract. Related: `test_fragment_repository_query_summary.py:57-63` builds a `TextLine` dict with no `lineNumber` by hand, which the real contract never produces, and it passes anyway.

#### I1 — Info — response size grew with full lines

Measured at runtime: one factory-built text line in the preview is about 6.3 KB. Worst case at the clamp is 200 items × 5 lines, roughly 6 MB of preview per page; at the frontend's page size (101) about 3 MB. That's the cost of shipping `content` (the F3 decision), and it's acceptable — but "compact previews" in the description's Sourcery summary no longer describes it.

#### I2 — Info — this needs to deploy together with frontend #787

Frontend master never sends `limit`, so the limited path — and every change in this PR's response shape — is only used once #787 lands. Coordinate the two.

#### I3 — Info — small leftovers

`fragment_query_bibliography.py:57` sets `requested = set(fetched)`, so a missing referenced id that is also some chain's redirect target is queried twice (harmless, one extra `$in` element). `mongo_fragment_repository_get.py:69-78` `query_museum_numbers` now returns a list instead of a cursor; no production caller.

PR #754, opened 2026-08-20
Bibliography identity management

Fabdulla1:
## Trusted bibliography identity management

Adds the deliberate path for changing bibliography identity state, replacing
what PR #752 parked when it made ordinary updates metadata-only.

**Depends on #752.** `master` does not contain it, so `fix-bib` is merged into
this branch. Merge #752 first.

### What

`POST /bibliography/{id}/identity`, guarded by a new `admin:bibliography` scope,
manages `aliases`, `citationKey`, `deprecated` and `redirectTo` on an existing
record. `POST /bibliography/{id}` stays metadata-only. Partner APIs are unchanged.

### Request contract

Explicit commands, not full intended state:

```json
{
  "addAliases": [{"value": "…", "normalizedValue": "…", "type": "…"}],
  "removeAliases": ["…"],
  "citationKey": "…" | null,
  "deprecateTo": "CANONICAL_ID",
  "reactivate": true
}



## Summary by Sourcery

Introduce a trusted, explicitly authorized workflow for changing bibliography identity while keeping ordinary and partner metadata updates identity-preserving.

New Features:
- Add a restricted `admin:bibliography` endpoint for explicitly managing aliases, citation keys, and deprecation redirects on existing bibliography records.

Bug Fixes:
- Prevent ordinary metadata updates from mutating or losing server-owned bibliography identity fields.
- Reject stale or concurrent bibliography updates with conflicts instead of silently overwriting identity changes.
- Preserve identity data and lookup behavior for legacy records and partner updates.

Enhancements:
- Validate deprecation targets against existing redirect rules and support reactivating or repairing tombstones.
- Add atomic lookup reservation handling, audit logging, failure recovery, and compare-and-set protection for identity changes.

Tests:
- Add comprehensive coverage for identity commands, authorization, collisions, redirects, legacy data, concurrency, recovery, and metadata compatibility.

Chores:
- Merge the bibliography update foundation required for identity management.

## Summary by Sourcery

Introduce an explicitly authorized workflow for managing bibliography identity while keeping existing metadata updates identity-preserving.

New Features:
- Add a restricted identity-management endpoint for modifying bibliography aliases, citation keys, deprecation redirects, and reactivation state.

Bug Fixes:
- Keep ordinary metadata and partner updates from changing or losing server-owned bibliography identity fields.
- Detect stale or concurrent identity changes as conflicts instead of overwriting them.
- Preserve lookup ownership and redirect behavior for legacy bibliography records.

Enhancements:
- Add validation for redirect chains, identity collisions, compare-and-set updates, atomic lookup reservations, recovery handling, and audit logging.

Documentation:
- Document the restricted bibliography administration scope and its provisioning and single-writer operational requirements.

Tests:
- Add comprehensive coverage for identity commands, authorization, metadata compatibility, redirects, collisions, legacy records, concurrency, recovery, and reservation behavior.

### CODE REVIEW

**Review by khoidt — CHANGES_REQUESTED** [review]
Really nice piece of work. The separation is the right one: `POST /bibliography/{id}` is metadata-only, identity moves through one explicit, admin-scoped command endpoint, and partner routes are untouched. I ran the real service against a local Mongo and went through it properly — aliases, citation keys, deprecation, reactivation, redirect chains, collisions, and the authorization boundaries all behave exactly as described. Every gate is green here: pyre, pyright, ruff, flake8, mypy, `task lint-md`, 4604 tests, and 100% coverage on all twenty changed modules. CI, CodeQL and qlty are clean too, and there are no devcontainer or CI-config changes anywhere on the branch.

Both of Codex's P1s are genuinely dealt with, not just marked resolved — I rebuilt the six-hop chain by hand and the inbound-depth check catches it with the right error naming the right entry. The cross-record race is mitigated with a rollback, honestly documented, and has a regression test. All ten qlty comments are gone.

One thing blocks merge outright: this PR adds a new `.md` file to the repository, `docs/handoffs/admin-bibliography-scope-provisioning.md`, with a `README.md` hunk pointing at it. New markdown files are not landing in this repo, so that has to come out before anything else happens — the content is good and belongs in the PR description or a ticket, not in the tree.

Two more I'd like sorted alongside it, neither of them deep. The create route is still the open half of the story: `write:bibliography` alone can mint a `citationKey` and `aliases` at birth, claiming values in the very namespace the new admin scope exists to protect — I did it on the running service. And #752 is still open with changes requested, while this branch already carries its commits, so the merge order needs a decision. The rest below is low-severity or a nit.

### Details

#### 1. [Blocking] A new `.md` file is added to the repository

`docs/handoffs/admin-bibliography-scope-provisioning.md` (115 lines) is added by this PR, and `README.md` gains a four-line Scopes note that links to it.

The document itself is good — it is candid about the scope being unprovisioned, about the single-writer constraint, and about what was fixed and why. But no new `.md` files should be landing in this repository, so it needs to go somewhere else: the PR description, a ticket, or the team handoff channel.

Note the coupling: the README hunk points at the file by path, so if the doc is deleted and the README note is kept, the link dangles. Both hunks need to be handled together.

**Severity:** Blocking. Not a code defect — a repository rule, and one that is not negotiable at review time. Everything else in this review can be argued or deferred; this cannot merge as it stands.
**Reproduction:** `git diff --name-status origin/master...HEAD -- '*.md'` → `M README.md`, `A docs/handoffs/admin-bibliography-scope-provisioning.md`.
**Recommendation:** Drop the new doc. Move the provisioning steps and the single-writer constraint into the PR description; keep a one-line scope entry in the README without the file reference, or drop the README hunk too.

#### 2. [Medium] The create route still lets `write:bibliography` claim identity values

`reject_born_identity_lifecycle_fields` blocks only `deprecated` and `redirectTo`. `CSL_JSON_SCHEMA` still accepts `citationKey` and `aliases` on `POST /bibliography`, and `create_with_identity_claims` duly reserves them.

So the field set that `POST /bibliography/{id}/identity` now guards behind `admin:bibliography` is writable at birth by any `write:bibliography` holder. Concretely, a non-admin can take `Smith2020` — or any alias, or any string in the shared lookup namespace — by creating a throwaway entry, and an admin then cannot assign that value to the correct record without first going to the squatted entry and removing it. The 409 is correct; the asymmetry is the problem.

I appreciate this is arguably deliberate — an entry does need a citation key at birth, and this half of it was parked in #752 as well. But the PR's stated model is "identity changes require `admin:bibliography`", and creation is a hole in exactly that.

**Severity:** Medium (authorization asymmetry; no data corruption, and collisions are still rejected).
**Reproduction (verified live, `127.0.0.1:8125`, scopes `write:bibliography` only for the create):**

```console
$ curl -X POST $B/bibliography -H 'Content-Type: application/json' \
    -d '{"id":"RS-10","type":"book","title":"T","citationKey":"Born2020","aliases":[{"value":"BornAlias"}]}'
201

$ curl -o /dev/null -w '%{http_code}\n' $B/bibliography/Born2020
200
```

Contrast with the same fields on the update route, which correctly returns 409, and with `deprecated`/`redirectTo` on create, which correctly return 422.

**Recommendation:** Either extend `reject_born_identity_lifecycle_fields` to cover `citationKey` and `aliases` (and let creation be metadata-only, with identity set by a follow-up identity call), or state explicitly in the PR description that birth-time identity is intentionally a `write:bibliography` capability and say why. Please do not leave it implicit.

#### 3. [Medium] Merge order: #752 is still open with changes requested

This branch carries two `Merge branch 'fix-bib'` commits, so #754's diff includes all of #752. #752 is still `OPEN`, and my most recent review of it (2026-08-25) is `CHANGES_REQUESTED`. The PR body says "Merge #752 first."

Worth knowing: the three gate failures I held #752 for are all fixed on this branch — pyright is clean on `test_bibliography_stale_identity_conflict.py`, flake8 reports no `B042` on `BibliographyUpdateConflictError`, and the stray `.devcontainer` reformatting is gone. So #754 is ahead of #752, not behind it.

**Severity:** Medium (process/merge safety, not code).
**Reproduction:** `git log --oneline --merges origin/master..HEAD`; `gh pr view 752 --json state,reviews`.
**Recommendation:** Merge #752 first as the body says, or close #752 in favour of this one and say so on that PR. Do not merge #754 while #752's review state still reads as blocking without a decision recorded.

#### 4. [Low] A failed rollback can leave a permanent A↔B redirect loop

`BibliographyIdentityManagement._reject_concurrent_redirect_break` re-validates after the write and, on failure, restores `stored_entry` with an identity-only compare-and-set, then raises `BibliographyUpdateConflictError`. That is a neat mitigation and it is covered by `test_a_concurrent_cross_record_redirect_break_is_rolled_back`.

The residual: the restore is itself CAS-guarded on the state this request just wrote. If a third writer changes that record between the write and the restore, `update_identity_fields_only` raises, and control never reaches the `raise BibliographyUpdateConflictError(entry["id"]) from error` line. The caller still gets a 409 (the inner error is also a `DuplicateError`), but the record is left deprecated toward a record that redirects back — both then unresolvable, with `GET` returning 409.

The handoff doc's limitation section mentions only "a genuine double-persist-before-either-re-check". This is a second residual shape, and it needs no simultaneity between the two deprecations — just a third identity write landing inside the rollback window.

**Severity:** Low (unreachable under the documented single-writer operating constraint).
**Reproduction:** Code path, not reproduced at runtime — `ebl/bibliography/application/identity_management.py:61-65`. Triggering it needs three concurrent identity writers, which the operating constraint forbids.
**Recommendation:** Wrap the restore so a failing rollback is logged loudly and still surfaces the intended conflict, and add this shape to the limitation list in whatever finally carries that documentation.

#### 5. [Low] Inbound-chain validation is linear in the number of inbound tombstones

`_validate_inbound_chains` walks every entry whose `redirectTo` points at the entry being deprecated, then their predecessors, following each forward. Depth is capped at `MAX_REDIRECT_DEPTH`, but fan-in is not capped at all — and a canonical entry that has absorbed many merged duplicates is precisely the shape that accumulates inbound tombstones.

Measured on the running service (the `redirectTo` index exists, so each query is cheap; the count is the issue):

| inbound predecessors | deprecate the hub |
| --- | --- |
| 0 | 16.5 ms |
| 50 | 103.4 ms |
| 200 | 361.4 ms |

That is roughly 1.7 ms per inbound tombstone, linear. At 2000 merged duplicates a single `deprecateTo` would sit around three and a half seconds inside a request that also holds pending lookup reservations.

**Severity:** Low (correct, and fine at today's data volumes).
**Reproduction:** Insert N tombstones with `redirectTo` set to a hub id, then `POST /bibliography/{hub}/identity {"deprecateTo": ...}` and time it.
**Recommendation:** No change needed now. Worth a note in the follow-up issue, since the natural fix (validating depth from the affected head only, or storing chain depth) arrives with the cross-record transaction work anyway.

#### 6. [Low] The identity route returns a locally computed entry, not a re-read

`manage_identity` returns `entry`, built from the `stored_entry` read at the start of the request. Because `update_identity_fields` writes only the four identity fields, a metadata edit landing between that read and the response is correctly preserved in the database but absent from the 200 body — the caller is told the entry still has the old title.

Only the response is stale; nothing is lost. Still, a curator tool that writes back what it just received would then send stale metadata into the metadata route.

**Severity:** Low.
**Reproduction:** Code path — `ebl/bibliography/application/identity_management.py:29-40`. Not reproduced at runtime; it needs an interleaving I could not force over plain HTTP.
**Recommendation:** Return `self._repository.query_by_id(id_)` after a successful write.

#### 7. [Low] `suppress(NotFoundError)` in `reconcile_reservation` covers more than intended

In `ebl/bibliography/infrastructure/lookup_reservation_reconciliation.py:34-38` the `with suppress(NotFoundError)` block wraps the `owns_value(entry_id, value)` call as well as the transition it guards. The idempotency fix itself is right and matches the existing `retire`/`release_pending` pattern.

`_entry_owns_lookup_value` cannot currently raise `NotFoundError`, so this is latent — but if it ever does, reconciliation will silently skip instead of failing.

**Severity:** Low (latent).
**Recommendation:** Hoist the `owns_value` call above the `with` block so only the transition is suppressed.

#### 8. [Nit] The redirect-loop message names the wrong end of the chain

Rejecting `RS-1 → RS-2` while `RS-2 → RS-1` already exists returns `Bibliography redirect loop detected at RS-2`. Correct rejection, but it names the record the curator did not touch. The depth message gets this right — it names `C-1`, the affected head.

**Recommendation:** Mention the entry being changed alongside the record where the loop closes.

#### 9. [Nit] The identity route accepts the partner alias sentinels

`BIBLIOGRAPHY_ALIAS_SCHEMA` allows free-form `type`, `source` and `status`, so an admin can add an alias carrying `partner_id` / `partner_request` / `redirect` — the exact values `create_partner_alias` stamps on server-generated partner aliases.

This is not a defect today: I grepped, and `PARTNER_ALIAS_TYPE`, `PARTNER_ALIAS_SOURCE` and `PARTNER_ALIAS_STATUS` are written but never read anywhere in production code, so nothing discriminates on them. Flagging it only because the alias schema's own docstring says "future tightening may require a migration" — if anything ever does start reading those fields to tell alias kinds apart, hand-written aliases from this route become indistinguishable from partner-owned ones.

**Recommendation:** Nothing now. If alias kinds ever become meaningful, they must be separated structurally rather than by probing `type`/`source`.

PR #755, opened 2026-08-20
fix: fix mongo not found errors

Fabdulla1:
## fix: sanitize Mongo not-found errors

Generic `MongoCollection` not-found errors previously interpolated the raw Mongo query into the HTTP `404` description. This exposed internal field names and Mongo predicates such as `_id`, `$or`, `$exists`, `citationKey`, and `aliases.value` to API clients.

### What changed

`MongoCollection` now distinguishes between:

* **identifier-based lookups**, where the repository API explicitly knows a single safe identifier and returns messages such as `bibliography clientChosenKey not found.`;
* **arbitrary query-based lookups**, which now return only a resource-level message such as `bibliography not found.`.

Raw Mongo query dictionaries are never rendered into public error responses.

Application/domain-specific `NotFoundError` messages remain unchanged, so useful errors such as `Fragment X.999 not found.` continue to be returned where the application layer knows the semantic identifier.

### Compatibility

This does not change:

* HTTP status codes;
* the Falcon error response shape;
* successful Mongo operations;
* domain-specific not-found messages;
* partner APIs or bibliography identity behavior.

Only the generic `404` description changes.

No repository code or tests were found to parse or depend on the previous raw-query error text.

### Scope

The issue was discovered through bibliography updates but existed in shared `MongoCollection` infrastructure and affected multiple domains, including bibliography, dictionary, signs, provenance, cache, corpus, fragmentarium, and other repositories using the same abstraction.

The old behavior was a low-severity information-disclosure/API-hygiene issue: it exposed query structure and internal schema details, but no confirmed server-held secrets or privileged data.

### Tests

Added coverage for:

* exact-ID lookups;
* complex `$or` queries;
* `$in`, `$exists`, `$ne`, `$elemMatch`, and other Mongo predicates;
* repository behavior across representative domains;
* HTTP response serialization;
* preservation of existing domain-specific error messages.

The existing `MongoCollection` test module was split to remain under the repository's 250-line file limit without removing or weakening any existing tests.

No production data, bibliography identity behavior, partner APIs, or unrelated feature branches are changed by this PR.

## Summary by Sourcery

Prevent MongoDB query details from leaking through generic not-found responses while retaining meaningful identifier and domain-specific errors.

Bug Fixes:
- Sanitize generic MongoDB not-found errors so HTTP 404 responses no longer expose raw query structures or internal field names.
- Preserve safe identifier-based and domain-specific not-found messages across repositories and routes.

Enhancements:
- Add explicit identifier-based Mongo collection operations with resource-and-identifier error messages while using resource-level messages for arbitrary queries.

Tests:
- Add coverage for sanitized Mongo errors across CRUD operations, repositories, representative routes, and domain-specific message preservation.
- Reorganize Mongo collection tests into focused modules without reducing existing coverage.

### DIFF
--- a/ebl/corpus/infrastructure/mongo_text_repository_get.py
+++ b/ebl/corpus/infrastructure/mongo_text_repository_get.py
@@ -3,22 +3,29 @@ from ebl.corpus.domain.chapter import ChapterId
 from ebl.corpus.infrastructure.queries import (
     chapter_id_query,
 )
-from ebl.corpus.infrastructure.mongo_text_repository_base import MongoTextRepositoryBase
+from ebl.corpus.infrastructure.mongo_text_repository_base import (
+    MongoTextRepositoryBase,
+    chapter_not_found,
+)
+from ebl.errors import NotFoundError
 
 
 class MongoTextRepositoryGet(MongoTextRepositoryBase):
     def get_sign_data(self, id_: ChapterId) -> dict:
-        return self._chapters.find_one(
-            chapter_id_query(id_),
-            projection={
-                "_id": False,
-                "signs": True,
-                "manuscripts": True,
-                "textId": True,
-                "stage": True,
-                "name": True,
-            },
-        )
+        try:
+            return self._chapters.find_one(
+                chapter_id_query(id_),
+                projection={
+                    "_id": False,
+                    "signs": True,
+                    "manuscripts": True,
+                    "textId": True,
+                    "stage": True,
+                    "name": True,
+                },
+            )
+        except NotFoundError as error:
+            raise chapter_not_found(id_) from error
 
     def get_all_sign_data(self) -> Sequence[dict]:
         return list(
--- a/ebl/dictionary/infrastructure/word_repository.py
+++ b/ebl/dictionary/infrastructure/word_repository.py
@@ -227,4 +227,4 @@ class MongoWordRepository(WordRepository):
         return self._collection.get_all_values("_id")
 
     def update(self, word) -> None:
-        self._collection.update_one({"_id": word["_id"]}, {"$set": word})
+        self._collection.update_one_by_id(word["_id"], {"$set": word})
--- a/ebl/fragmentarium/infrastructure/mongo_fragment_repository_get.py
+++ b/ebl/fragmentarium/infrastructure/mongo_fragment_repository_get.py
@@ -405,17 +405,14 @@ class MongoFragmentRepositoryGetBase(MongoFragmentRepositoryBase):
     def query_next_and_previous_fragment(
         self, museum_number: MuseumNumber
     ) -> FragmentPagerInfo:
-        try:
-            current = self._fragments.find_one(
-                {
-                    "museumNumber.prefix": museum_number.prefix,
-                    "museumNumber.number": museum_number.number,
-                    "museumNumber.suffix": museum_number.suffix,
-                },
-                projection={"_sortKey": True},
-            ).get("_sortKey")
-        except NotFoundError as error:
-            raise NotFoundError(f"Fragment {museum_number} not found.") from error
+        current = self._fragments.find_one(
+            {
+                "museumNumber.prefix": museum_number.prefix,
+                "museumNumber.number": museum_number.number,
+                "museumNumber.suffix": museum_number.suffix,
+            },
+            projection={"_sortKey": True},
+        ).get("_sortKey")
 
         if current is None:
             prev = next_ = museum_number
--- a/ebl/fragmentarium/infrastructure/mongo_fragment_repository_get.py
+++ b/ebl/fragmentarium/infrastructure/mongo_fragment_repository_get.py
@@ -405,14 +405,17 @@ class MongoFragmentRepositoryGetBase(MongoFragmentRepositoryBase):
     def query_next_and_previous_fragment(
         self, museum_number: MuseumNumber
     ) -> FragmentPagerInfo:
-        current = self._fragments.find_one(
-            {
-                "museumNumber.prefix": museum_number.prefix,
-                "museumNumber.number": museum_number.number,
-                "museumNumber.suffix": museum_number.suffix,
-            },
-            projection={"_sortKey": True},
-        ).get("_sortKey")
+        try:
+            current = self._fragments.find_one(
+                {
+                    "museumNumber.prefix": museum_number.prefix,
+                    "museumNumber.number": museum_number.number,
+                    "museumNumber.suffix": museum_number.suffix,
+                },
+                projection={"_sortKey": True},
+            ).get("_sortKey")
+        except NotFoundError as error:
+            raise NotFoundError(f"Fragment {museum_number} not found.") from error
 
         if current is None:
             prev = next_ = museum_number
--- a/ebl/mongo_collection.py
+++ b/ebl/mongo_collection.py
@@ -44,13 +44,18 @@ class MongoCollection:
         raise AssertionError("insert_one should return or raise")
 
     def find_one_by_id(self, id_):
-        return self.find_one({"_id": id_})
+        document = self.__get_collection().find_one({"_id": id_})
+
+        if document is None:
+            raise self.__id_not_found_error(id_)
+        else:
+            return document
 
     def find_one(self, query, *args, **kwargs) -> Any:
         document = self.__get_collection().find_one(query, *args, **kwargs)
 
         if document is None:
-            raise self.__not_found_error(query)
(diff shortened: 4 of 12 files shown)


khoidt (approved):
Nice one — this is a clean, well-scoped fix. Raw Mongo queries no longer end up in `404` bodies, the identifier-based messages that people actually rely on are kept, and I confirmed both on a running instance rather than just from the tests. Splitting the old 262-line test module also quietly fixed a file that was already over the 250-line limit, and every original test survived the move. CI is fully green, qlty reports no blocking issues, and Sourcery had nothing to flag.

I found ten things worth mentioning, all minor — nothing that should hold up the merge. The two I'd actually act on before merging are the small naming fixes in the tests (a constant called `SECRET_VALUE` that the test asserts *is* echoed back, and a test named `..._keeps_domain_message` that asserts a generic message). The rest are follow-ups: `update_one`/`delete_one` still drop identifiers they legitimately know, internal collection names are still rendered to clients, and the dropped query is now lost for diagnostics too since nothing logs it. **Verdict: approve, with the two test-naming nits worth a quick follow-up commit.**

### Details

#### 1. `update_one` / `delete_one` discard an identifier that is known and safe — Low

The PR's stated principle is to keep the identifier "where the repository API explicitly knows a single safe identifier". Four call sites do know one, yet now produce a bare resource message because they go through the generic helpers:

| call site | query | message after this PR |
| --- | --- | --- |
| [`word_repository.py:230`](ebl/dictionary/infrastructure/word_repository.py#L230) | `{"_id": word["_id"]}` | `word not found.` |
| [`mongo_text_repository_modify.py:79`](ebl/corpus/infrastructure/mongo_text_repository_modify.py#L79) | `chapter_id_query(id_)` | `chapter not found.` |
| [`mongo_fragment_repository.py:123`](ebl/fragmentarium/infrastructure/mongo_fragment_repository.py#L123) | `fragment_is(fragment)` | `fragment not found.` |
| [`lookup_reservations.py:226`](ebl/bibliography/infrastructure/lookup_reservations.py#L226) | `{"_id": value, "state": …}` | `bibliography_lookup_reservation not found.` |

`find_one_by_id` and `replace_one` keep the identifier in exactly this situation, so the treatment is inconsistent within one class.

I initially rated this Medium and probed it on the live service. It is **not** reachable on the normal write paths: [`dictionary_service.py:103`](ebl/dictionary/application/dictionary_service.py#L103) reads the word first, and [`corpus.py:197`](ebl/corpus/application/corpus.py#L197) reads the chapter first, so the read raises the identifier-bearing error before the write is attempted. `POST /words/missing%20word%20I` on the running service returns `word missing word I not found.`, not the bare message. The bare message therefore only surfaces on a read-then-write race — hence Low.

Suggested fix: add `update_one_by_id(id_, update)` / `delete_one_by_id(id_)` that route through `__id_not_found_error`, or give the generic helpers an optional `identifier` argument.

#### 2. Internal collection names still reach the client — Low (residual information disclosure)

`__resource_noun` is derived from the Mongo collection name and is still rendered into the public description. Field names and predicates are gone; collection names are not. Verified with `singlar()`:

| collection | rendered noun in the 404 body |
| --- | --- |
| `bibliography_lookup_reservations` | `bibliography_lookup_reservation` |
| `cropped_sign_images` | `cropped_sign_image` |
| `cache` | `cache` |
| `changelog` | `changelog` |

For `words`/`fragments`/`bibliography` the noun is the user-facing resource name and is entirely appropriate. For internal collections it is internal schema, which is the category this PR set out to stop leaking. `ebl/tests/mongo/test_repository_not_found.py:52-55` even asserts `cache not found.` as expected behaviour. Either map internal collections to a neutral public noun, or state explicitly that collection names are considered public.

#### 3. The dropped query is now lost for diagnostics too — Low

[`ebl/error_handler.py:24-25`](ebl/error_handler.py#L24-L25) is `raise falcon.HTTPNotFound(description=str(ex))` and logs nothing; `NotFoundError` in [`ebl/errors.py:1-2`](ebl/errors.py#L1-L2) is a bare `Exception` with no structured payload. Before this PR the query at least appeared in the Sentry event title; now every generic not-found on a collection produces the identical string, so Sentry can group them only by stack frame and the failing values are gone entirely. Sanitising the *response* is right; discarding the value everywhere is a side effect worth avoiding. Suggested fix: `NotFoundError(message, query=query)` with only `message` rendered, or a `logging.debug` inside `__not_found_error`.

#### 4. `find_one_by_id` no longer delegates to `find_one` — Low (maintainability)

[`ebl/mongo_collection.py:46-52`](ebl/mongo_collection.py#L46-L52) now calls `__get_collection().find_one(...)` directly and repeats the `document is None` check from `find_one` four lines below. I grepped for subclasses and overrides — there are none (`MongoCollection` is only ever instantiated, never subclassed), so there is no behavioural risk today. The two lookups can drift, though; a shared private `__find_one_or_none(query)` used by both would keep them in step.

#### 5. `find_one_by_id` is a modified function with no type hints — Low (coding standard)

`def find_one_by_id(self, id_):` has neither a parameter nor a return annotation, while the new `__id_not_found_error(self, id_: object) -> NotFoundError` immediately below is fully annotated. The project standard requires type hints on all functions and methods, and this one was touched by the PR. `def find_one_by_id(self, id_: object) -> Any:` matches the sibling `find_one`.

#### 6. `ebl/mongo_collection.py` sits at 98% coverage — Low

`poetry run pytest ebl/tests/mongo --cov=ebl.mongo_collection --cov-report=term-missing` gives 95 statements, 2 missed, **98%**, missing lines **44** and **102**. Both are `raise AssertionError(...)` guards after `for attempt in range(2)` loops that always return or raise, so they are unreachable by construction. qlty reports 100% diff coverage because the lines this PR *edited* are covered — line 102 nonetheless sits inside `update_one`, whose body the PR modified at line 100. Either restructure the retry loop so there is no unreachable tail, or mark the two guards `# pragma: no cover`.

#### 7. Misleading test name: `test_bibliography_get_of_missing_entry_keeps_domain_message` — Low

[`ebl/tests/mongo/test_not_found_routes.py:35-38`](ebl/tests/mongo/test_not_found_routes.py#L35-L38) asserts `bibliography clientChosenKey not found.`, which is the **generic** identifier message from `MongoCollection.find_one_by_id` — `MongoBibliographyRepository.query_by_id` is a one-line delegation to it ([`bibliography.py:77-79`](ebl/bibliography/infrastructure/bibliography.py#L77-L79)). There is no domain message on that path. Rename to `test_bibliography_get_of_missing_entry_reports_resource_and_identifier`, otherwise a future refactor that genuinely removes a domain message will not be caught by the test that claims to guard it.

#### 8. `SECRET_VALUE` names a value the test asserts *is* echoed back — Low

[`ebl/tests/mongo/test_mongo_collection_not_found.py:6`](ebl/tests/mongo/test_mongo_collection_not_found.py#L6) defines `SECRET_VALUE = "secret-or-user-supplied-value"`, and three tests (`test_find_one_by_id_reports_resource_and_identifier`, `test_replace_one_reports_resource_and_identifier`, `test_replace_one_with_filter_reports_document_identifier`) assert that it **appears** in the message. In a file whose subject is preventing leaks, a constant called `SECRET_VALUE` that is deliberately reflected reads as a bug. Rename to `IDENTIFIER`.

#### 9. The "no `{` / `$` / `_id`" assertion is reimplemented three times — Low (reuse)

`assert_no_query_details` ([`test_mongo_collection_not_found.py:18`](ebl/tests/mongo/test_mongo_collection_not_found.py#L18)), `assert_message` ([`test_repository_not_found.py:9`](ebl/tests/mongo/test_repository_not_found.py#L9)) and `assert_sanitized` ([`test_not_found_routes.py:15`](ebl/tests/mongo/test_not_found_routes.py#L15)) each re-derive the same forbidden-substring check with slightly different lists. One helper in `ebl/tests/mongo/conftest.py` would keep the three in agreement as the list grows.

#### 10. No repository- or route-level coverage of the sanitized write paths — Low (test coverage)

`test_repository_not_found.py` exercises only `find_one` (provenance long name, cache get). `update_one`, `delete_one` and `delete_many` are covered at the `MongoCollection` level but never through a repository or a route, which is where a regression would actually reach a client. One repository-level test on the cache delete path would close it.

#### Informational: API-wide message style is still inconsistent

Live responses from the running service: `word missing-word not found.`, `Fragment X.999 not found.`, `Realia entry 'missing' not found.`, `bibliography clientChosenKey not found.` — three different casings and shapes. Pre-existing and out of scope, but this PR standardises the `MongoCollection` layer and is the natural place to note the follow-up.

PR #760, opened 2026-09-04
feat(map): add optional findspot map-location contract

Fabdulla1:
## Stack and cross-repository dependencies

- **Direct same-repository parent:** none — this PR targets `master` and is a root of the backend map work.
- **Current base branch:** `master`
- **Backend/frontend contract dependency:** none directly. The `MapLocation` field vocabulary reaches the frontend only through the `/findspots/map-data` response shape delivered by #759.
- **Dependency type:** none in isolation — internal domain and schema foundation.
- **Recommended merge and deployment order:** may merge on its own once approved and checked; no cross-repository coordination is required for this PR by itself. Required checks and human approvals are separate from dependency order.
- **Direct downstream active PR:** none by Git ancestry — see the branch-state note below.

### Branch state — read before merging

This branch is **not** an ancestor of `map/backend-02-findspot-query` (#761). However, `ebl/fragmentarium/domain/map_location.py` and `ebl/fragmentarium/application/map_location_schema.py` are **byte-identical** on both branches, so #761 already carries this PR's contract content independently.

The `map/backend-02..06` branches were cut from this branch's original commit and do not contain its two later commits (`test(map): close coverage/type gaps…`, `fix(map): normalize MapLocation source…`). Reviewers should decide how to reconcile the duplication — rebasing #761 onto this branch's tip, or merging this first and rebasing the rest — before merging both. This branch is also behind `master`. No rebase or retarget has been performed as part of this description update.

---

## What this PR introduces

- `MapLocation` domain model on `Findspot` — optional, with `polygon_ids`, `location_precision`, `match_method`, `source` and `source_revision`.
- `MapLocationPrecision` (`excavation-area`) and `MapLocationMatchMethod` (`curated`, `verified-source`) enumerations.
- Domain and schema validation and normalization: `polygonIds` must be non-empty, unique and free of blank values; `source` and `sourceRevision` are stripped and must be non-empty.
- Exposure of `mapLocation` through archaeology and fragment responses when present.
- Tests for domain validation, schema round-trips, persistence and fragment-API exposure.

This is a **backend-internal foundation**. It adds no HTTP endpoint and no query parameter.

## Frontend consumers

- **No direct runtime consumer.** No frontend PR imports or calls anything added here on its own.
- **Indirect only:** the vocabulary defined here surfaces to `ElectronicBabylonianLiterature/ebl-frontend#808` through the #759 endpoint response, whose rows carry `locationPrecision`, `matchMethod`, `polygonIds` and the `sector` / `area` / `building` / `room` fields. The frontend's `FindspotMapDataDto` mirrors that vocabulary but never depends on this PR in isolation.

Note on the `matchMethod` vocabulary: both `curated` and `verified-source` are valid in this contract, but **every mapping row committed in #759 and #758 uses `verified-source`**. No committed artifact currently uses `curated`.

## Database side effects

None. This PR adds no index declaration, no migration and no write path. The `archaeology.findspotId` index belongs to #761.

## Validation

No runtime verification claim is made here. The model, enumerations and validation rules above were read from the source at this PR's head.

## Summary by Sourcery

Add an optional, validated map-location model and response contract for findspots.

New Features:
- Add an optional map-location contract to findspots, including polygon identifiers, precision, matching method, and source metadata.
- Expose map-location data through findspot and fragment archaeology responses when available.

Enhancements:
- Introduce domain and schema validation with normalization for map-location values and supported enumerations.

Tests:
- Add coverage for map-location validation, schema round-trips, persistence, and fragment API exposure.

### DIFF
--- a/ebl/fragmentarium/application/archaeology_schemas.py
+++ b/ebl/fragmentarium/application/archaeology_schemas.py
@@ -34,7 +34,7 @@ class ExcavationPlanSchema(Schema):
         return ExcavationPlan(**data)
 
 
-class ProvenanceSiteMixin:
+class ProvenanceSiteMixin(Schema):
     def serialize_site(self, obj):
         return getattr(obj.site, "long_name", None)
 
--- a/ebl/fragmentarium/application/archaeology_schemas.py
+++ b/ebl/fragmentarium/application/archaeology_schemas.py
@@ -7,6 +7,7 @@ from ebl.fragmentarium.domain.archaeology import (
     Archaeology,
     ExcavationNumber,
 )
+from ebl.fragmentarium.application.map_location_schema import MapLocationSchema
 from ebl.fragmentarium.domain.findspot import (
     BuildingType,
     ExcavationPlan,
@@ -14,7 +15,7 @@ from ebl.fragmentarium.domain.findspot import (
 )
 from ebl.common.application.schemas import deserialize_provenance_record
 from ebl.schemas import NameEnumField
-from marshmallow import Schema, fields, post_load
+from marshmallow import Schema, fields, post_dump, post_load
 
 
 class ExcavationNumberSchema(AbstractMuseumNumberSchema):
@@ -55,6 +56,9 @@ class FindspotSchema(ProvenanceSiteMixin, Schema):
     plans = fields.Nested(ExcavationPlanSchema, many=True, load_default=())
     room = fields.String()
     context = fields.String()
+    map_location = fields.Nested(
+        MapLocationSchema, allow_none=True, load_default=None, data_key="mapLocation"
+    )
     primary_context = fields.Boolean(
         data_key="primaryContext",
         allow_none=True,
@@ -65,6 +69,12 @@ class FindspotSchema(ProvenanceSiteMixin, Schema):
     def create_findspot(self, data, **kwargs) -> Findspot:
         data["plans"] = tuple(data["plans"])
         return Findspot(**data)
+
+    @post_dump
+    def omit_missing_map_location(self, data, **kwargs):
+        if data.get("mapLocation") is None:
+            data.pop("mapLocation", None)
+        return data
 
 
 class ArchaeologySchema(ProvenanceSiteMixin, Schema):
--- a/ebl/fragmentarium/application/map_location_schema.py
+++ b/ebl/fragmentarium/application/map_location_schema.py
@@ -1,0 +1,66 @@
+from marshmallow import (
+    EXCLUDE,
+    Schema,
+    ValidationError,
+    fields,
+    post_load,
+    validates_schema,
+    validate,
+)
+
+from ebl.fragmentarium.domain.map_location import (
+    MapLocation,
+    MapLocationMatchMethod,
+    MapLocationPrecision,
+)
+from ebl.schemas import ValueEnumField
+
+
+def _strip_or_none(value: str | None) -> str | None:
+    if value is None:
+        return None
+    stripped = value.strip()
+    return stripped or None
+
+
+class MapLocationSchema(Schema):
+    class Meta:
+        unknown = EXCLUDE
+
+    polygon_ids = fields.List(
+        fields.String(validate=validate.Length(min=1)),
+        required=True,
+        data_key="polygonIds",
+        validate=validate.Length(min=1),
+    )
+    location_precision = ValueEnumField(
+        MapLocationPrecision, required=True, data_key="locationPrecision"
+    )
+    match_method = ValueEnumField(
+        MapLocationMatchMethod, required=True, data_key="matchMethod"
+    )
+    source = fields.String(required=True, validate=validate.Length(min=1))
+    source_revision = fields.String(
+        required=True, data_key="sourceRevision", validate=validate.Length(min=1)
+    )
+
+    @validates_schema
+    def validate_map_location(self, data, **kwargs) -> None:
+        polygon_ids = data.get("polygon_ids", ())
+        if len(set(polygon_ids)) != len(polygon_ids):
+            raise ValidationError("polygonIds must be unique.", "polygonIds")
+        if any(not polygon_id.strip() for polygon_id in polygon_ids):
+            raise ValidationError(
+                "polygonIds must not contain empty values.", "polygonIds"
+            )
+        if _strip_or_none(data.get("source")) is None:
+            raise ValidationError("source must not be empty.", "source")
+        if _strip_or_none(data.get("source_revision")) is None:
+            raise ValidationError("sourceRevision must not be empty.", "sourceRevision")
+
+    @post_load
+    def create_map_location(self, data, **kwargs) -> MapLocation:
+        data["polygon_ids"] = tuple(data["polygon_ids"])
+        data["source"] = data["source"].strip()
+        data["source_revision"] = data["source_revision"].strip()
+        return MapLocation(**data)
--- a/ebl/fragmentarium/application/map_location_schema.py
+++ b/ebl/fragmentarium/application/map_location_schema.py
@@ -1,5 +1,4 @@
 from marshmallow import (
-    EXCLUDE,
     Schema,
     ValidationError,
     fields,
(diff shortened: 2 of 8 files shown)


khoidt (changes requested):
Nice, tight foundation PR. The `MapLocation` value object is small and immutable, the enums are right, the schema rejects unknown keys, and the read path does exactly what it says: I ran the service against a local database and `mapLocation` shows up on `/findspots` and on `GET /fragments/{n}` when it's set, and is left out when it isn't. The Codex point about stripping only on read is properly fixed — normalisation now lives in the domain. Every gate I ran is green and coverage is 100% on everything touched.

Two things need changing before I can approve, and both only show up when you actually drive the running service. First, the description says this adds no write path, but `POST /fragments/{n}/archaeology` now accepts `mapLocation` inside the embedded findspot and quietly stores it in the fragment document, where nothing ever reads it back. Second, a bad `mapLocation` on that same route is a 500, not a 422 — that route has never turned validation errors into 422s, but this PR adds eight new ways to hit it.

The rest is small: polygon IDs aren't trimmed the way `source` is, a behaviour change rode in on the merge commit, and the description's "byte-identical with #761" note has gone stale — the two branches now conflict.

No new `.md` files, and no dev container, Docker, CI, dependency or config changes — nothing to flag there.

### Details

#### M1 — Medium — the archaeology write route now accepts and persists `mapLocation`, which nothing ever reads back

`ebl/fragmentarium/application/archaeology_schemas.py:59-61` (the new nested field), reached through `ArchaeologySchema.findspot` at `:94` and `ebl/fragmentarium/web/archaeology.py:41-56`.

The PR description says: *"This PR adds no index declaration, no migration and no write path."* That is true for the `findspots` collection, but `FindspotSchema` is also the nested schema that `ArchaeologySchema` loads on `POST /fragments/{n}/archaeology`, and the updater stores the whole `archaeology` object — including the embedded `findspot` — into the fragment document.

Verified against the running service. A `transliterate:fragments` POST to `/fragments/X.2/archaeology` whose `findspot` carries `mapLocation: {polygonIds: ["client-1"], ...}` returns **200**, and the fragment document afterwards contains:

```text
fragments.archaeology.findspot.mapLocation = {'polygonIds': ['client-1'], 'locationPrecision': 'excavation-area', 'matchMethod': 'curated', 'source': 'client', 'sourceRevision': 'r1'}
findspots[4343].mapLocation               = <absent>
```

A re-`GET /fragments/X.2` shows no `mapLocation` at all, because `join_findspots` (`ebl/fragmentarium/infrastructure/queries.py:375-387`) overwrites `archaeology.findspot` with the `findspots` lookup. So the client's value is accepted, stored, and invisible — dead data in the fragment document that silently disagrees with the curated source of truth.

Today's frontend doesn't trigger this — `fromFindspotDto` builds `Findspot` positionally and drops unknown keys — but any API client can, and a frontend that starts carrying `mapLocation` on its `Findspot` (as the map work is heading towards) would echo it back on every archaeology save.

Fix: make `mapLocation` unacceptable on the archaeology write path rather than silently stored — e.g. load the request's nested findspot with a schema that excludes `map_location` so the existing `unknown=RAISE` turns it into a 422 (once M2 is fixed), and add a route test that posts `mapLocation` and asserts the rejection. Then the description's "no write path" is actually true.

#### M2 — Medium — an invalid `mapLocation` on the archaeology route is a 500, not a 422

`ebl/fragmentarium/web/archaeology.py:41-56` (root cause, pre-existing), newly reachable through `ebl/fragmentarium/application/map_location_schema.py:37-48`.

Verified against the running service, same route and token as M1:

| Payload | Response |
| --- | --- |
| `findspot.mapLocation.polygonIds = ["dup", "dup"]` | **500** — `ValidationError: {'findspot': {'mapLocation': {'polygonIds': ['polygonIds must be unique.']}}}` |
| `findspot.mapLocation.bogus = 1` | **500** — `ValidationError: {'findspot': {'mapLocation': {'bogus': ['Unknown field.']}}}` |
| `findspot.buildingType = "NOT_A_TYPE"` (pre-existing field, control) | **500** |

The control row shows the root cause predates this PR: `ArchaeologyResource.on_post` calls `ArchaeologySchema(...).load(data)` directly, and a marshmallow `ValidationError` is not mapped to 422 anywhere, so it falls through to the generic 500 handler. But this PR adds eight new rejection rules on that path (empty / duplicate / blank polygon IDs, bad precision, bad match method, blank source, blank revision, unknown key), and commit `3b8ce073` states *"drop `unknown = EXCLUDE` so unknown keys 422 like FindspotSchema"* — through the real route they don't, they 500.

Fix at the root: catch `ValidationError` in `ArchaeologyResource.on_post` and re-raise it as `DataError`, exactly as the sibling resources already do (`fragment_date.py:34-35`, `fragment_script.py:31-32`, `named_entities.py:74-75`); `error_handler.py:39` maps `DataError` to 422. Add a route test for an invalid `mapLocation` asserting 422. One small change covers every field on that route, not only the new ones.

#### L1 — Low — polygon IDs are not normalised, so padded duplicates pass the uniqueness rule

`ebl/fragmentarium/domain/map_location.py:14-17, 36-39`; `ebl/fragmentarium/application/map_location_schema.py:39-45`.

`source` and `source_revision` are stripped by a converter, but polygon IDs are only *checked* for blankness, never trimmed:

```text
MapLocation((' assur-1 ', 'assur-1'), ...).polygon_ids        -> (' assur-1 ', 'assur-1')   # accepted as two distinct ids
MapLocationSchema().load({'polygonIds': [' assur-1', 'assur-1'], ...})   -> (' assur-1', 'assur-1')
MapLocation((1,), ...)                                          -> AttributeError: 'int' object has no attribute 'strip'
```

Polygon IDs are join keys into the map artefacts, so a padded ID silently fails to match anything, and the uniqueness rule doesn't catch the near-duplicate. The non-string case raises `AttributeError` instead of the `ValueError` every other invariant uses. Fix: have `_to_tuple` strip each element and require `str` members (raise `ValueError` otherwise), and add the two cases to `test_map_location_rejects_invalid_values`.

#### L2 — Low — a behaviour change is hidden inside the merge commit

`1deb019b` "Merge master and harden map location polygon IDs".

`git show --remerge-diff 1deb019b` shows the merge carries its own edits: `_to_tuple` changes from `Iterable[str]` to a `list | tuple` guard that raises `ValueError`, plus two new test cases. The change itself is fine, but a code change inside a merge commit doesn't show up in `git log -p`, in bisect, or in "changes since last review", and it's exactly what made the "byte-identical" claim in L3 untrue. Please keep merges pure and put follow-up edits in their own commit going forward.

#### L3 — Low — the description's branch-state note is stale; #760 and #761 now conflict

PR description, "Branch state — read before merging".

The description says `map_location.py` and `map_location_schema.py` are byte-identical on this branch and on `map/backend-02-findspot-query` (#761). At the current heads that is no longer true for `map_location.py` (#761 still has the `Iterable` version of `_to_tuple`), and `git merge-tree` of the two heads reports:

```text
CONFLICT (content): Merge conflict in ebl/fragmentarium/domain/map_location.py
CONFLICT (add/add): Merge conflict in ebl/tests/fragmentarium/test_map_location.py
```

`final-map` (#753) differs in both files as well. Whichever of #760 / #761 lands second will need a hand merge, and the stack below #761 (#762 → #759 → #758 → #757) inherits it. Please update the description and pick the reconciliation now (merge this first and rebase #761 onto it, as the description itself suggests), rather than discovering it at merge time.

#### I1 — Info — schema and domain enforce the same four rules independently

`map_location_schema.py:37-48` and `map_location.py:34-42` each implement non-empty / unique / no-blank polygon IDs and non-blank source/revision, with separate messages. They agree today, but two copies of an invariant drift. An option is to let the domain be the single authority and convert its `ValueError` to a `ValidationError` in `create_map_location`; not required.

#### I2 — Info — small test nits

`ebl/tests/fragmentarium/test_findspot_repository_map_location.py:16-28`: `attr.evolve(FindspotFactory.build(...))` with no changes is a no-op wrapper, and `find_all()[0]` relies on the findspot being the only one in the collection — `find_all()` filtered by `id_` would be sturdier. The new tests pass `context=` to marshmallow schemas and so emit the same `RemovedInMarshmallow4Warning` as the existing findspot tests; consistent with the codebase, just noting it.

#### I3 — Info — `ProvenanceSiteMixin` now subclasses `Schema`

`archaeology_schemas.py:37`. Commit `3b8ce073` explains why (pyright, `self.context`), and it's correct. `FindspotSchema(ProvenanceSiteMixin, Schema)` and `ArchaeologySchema(ProvenanceSiteMixin, Schema)` now list `Schema` redundantly; harmless, could be dropped.

